PR #752, opened 2025-07-03
Add resource requests and limits

gileri:
# Context: 
Add default k8s resource requests and limits in chart to allow for [higher QoS class](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos).

### DIFF
--- a/chart/templates/minio.yaml
+++ b/chart/templates/minio.yaml
@@ -33,6 +33,19 @@ spec:
         resources:
           requests:
             storage: {{ .Values.minio.storage.diskSize }}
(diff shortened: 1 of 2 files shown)


jbanety (approved):
What are the implications of this PR?

gileri:
Thanks for the review @jbanety !

This will allow setting CPU/RAM resource requests and limits in deployments, elevating the Kubernetes QoS class, and make them less likely to be evicted under resource constraints.

But in case of memory leaks, it can lead to the pods being killed if it becomes out-of-memory.

PR #776, opened 2025-07-16
[backend/frontend] feat(enrollment/chunk6): display enrollment details

jbanety:
# Context: 

# How to test:  

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local tests

# Additional information:

Related #issueNumber


portal-api/src/modules/services/enrollment/enrollment.domain.ts
```diff
@@ -102,4 +110,42 @@ export const enrollmentDomain = {
...
+        'Subscription.service_instance_id',
+        '=',
+        'ServiceInstance.id'
+      )
+      .where(
+        'ServiceDefinition.identifier',
+        '=',
+        ServiceDefinitionIdentifier.OctiEnrollment
+      )
+      .where('Subscription.organization_id', '=', userSelectedOrganization)
```
hervyt, line 144: IMO, this type of security should be add in the securityLayer. WDYT @jpkha  ?

portal-api/src/modules/services/enrollment/enrollment.domain.ts
```diff
@@ -102,4 +110,42 @@ export const enrollmentDomain = {
...
+      .where('Subscription.status', '=', 'ACCEPTED')
+      .whereIn('Subscription.joining', ['SELF_JOIN', 'AUTO_JOIN'])
+      .select(['Service_Configuration.config']);
```
hervyt, line 147: .secureQuery()

portal-api/src/modules/services/enrollment/enrollment.resolver.ts
```diff
@@ -32,6 +32,9 @@ const resolvers: Resolvers = {
         });
       }
     },
+    octiInstances: async (_, _z, context) => {
+      return enrollmentApp.octiInstances(context);
```
hervyt, line 36: Why don't you export directly the octiInstances function (as we usually do?)

portal-front/schema.graphql
```diff
@@ -335,15 +336,23 @@ type DocumentConnection {
...
+  url: String!
+  title: String!
+  contract: String!
+}
+
+input OCTIPlatformInput {
+  id: ID!
   url: String!
   title: String!
+  contract: String!
```
hervyt, line 350: **nitpick**: we can be a bit more strict and type it as 'EE' or 'NotEE' only ? So that we are more typed (and we will have the Enum in the frontend for whatever we will need)

portal-api/src/modules/services/enrollment/enrollment.app.ts
```diff
@@ -19,6 +20,17 @@ import {
 } from './enrollment.domain';
 
 export const enrollmentApp = {
+  octiInstances: async (context: PortalContext): Promise<OctiPlatform[]> => {
+    const instances = await enrollmentDomain.octiInstances(context);
+    return instances.map((instance) => ({
+      __typename: 'OCTIPlatform',
```
hervyt, line 26: If it's correctly typed, you don't need this here, no ? 
And the typename should be the name of the table in our database... ?

portal-front/src/components/service/home/hooks/useServiceQueries.ts
```diff
@@ -46,6 +53,24 @@ const getOwnedServices = (queryRef: PreloadedQuery<userServiceOwnedQuery>) => {
...
+    queryRef
+  );
+
+  const [data] = useRefetchableFragment<
+    enrollOCTIInstancesQuery,
+    enrollOCTIInstanceListFragment$key
+  >(enrollOCTIInstanceListFragment, queryData);
+
+  return (
+    (data.octiInstances as unknown as enrollOCTIInstanceFragment$data[]) ?? []
```
hervyt, line 70: Why do you need to cast it here ?

portal-front/src/components/service/service-instance-card.tsx
```diff
@@ -83,13 +82,34 @@ const ServiceInstanceCard: React.FunctionComponent<
...
-            {isLinkService && (
+            {isExternalService(serviceInstance) && (
               <ArrowOutwardIcon className="ml-auto size-3 shrink-0" />
             )}
           </div>
-          <p className="txt-sub-content text-muted-foreground">
-            {serviceInstance.description}
-          </p>
+          {(isEnrollmentService(serviceInstance) && (
+            <div className="grid grid-cols-3 gap-s">
```
hervyt, line 90: There is a lot of div 🌵
You can't have it in only one div and play with css to display it the way you want ?

portal-front/messages/en.json
```diff
@@ -331,7 +334,6 @@
         "ExistingCsvFeedFile": "Existing Csv Feed file: {file_name}",
         "UpdateJSONFile": "Update JSON file",
         "DescriptionUpdateJSONFile": "You're about to upload a new json file. The previous json file will be deleted. Do you want to continue?",
-        "CsvFeedIllustration": "Illustration",
```
hervyt, line 334: Why did you remove all these keys ? Unused ? They're not remove in the fr.json

portal-front/messages/fr.json
```diff
@@ -160,6 +160,11 @@
         "Title": "Vous êtes sur le point d'enrôler votre instance OpenCTI dans XTM Hub",
         "Description": "Dans quel espace souhaitez-vous enrôler votre instance OpenCTI ?"
       }
+    },
+    "Details": {
+      "InstanceID": "ID de l'instance",
+      "InstanceURL": "URL de l'instance",
+      "Contract": "Contrat"
```
hervyt, line 167: You miss keys here compared to the en.json. 
Do you use translation plugin to check you translations files have same keys ?

portal-api/src/modules/services/enrollment/enrollment.domain.ts
jbanety: IMO this is more about expected logic than security.
We want to display the services for the current organization.

portal-front/schema.graphql
jbanety: I left it specially in string to be able to manage the different cases afterwards like the trials,....

portal-front/src/components/service/home/hooks/useServiceQueries.ts
jbanety: Because of the way fragments are handled by Relay. The TS typing isn't the real runtime type.

portal-api/src/modules/services/enrollment/enrollment.app.ts
hervyt: Yes, so if it does not exist, it should not be this typename... :o

jbanety: Typename is related to GraphQL.
It is OK to have GraphQL types not synced with the DB models.

portal-api/src/modules/services/enrollment/enrollment.domain.ts
jpkha: Actually, I would say both. Logically, we want to return only the selectedOrganization, but when accessing the ServiceInstance, we also need to ensure that access is restricted to the scope of the user's selected organization.

portal-api/src/modules/services/enrollment/enrollment.graphql
```diff
@@ -1,12 +1,20 @@
-input OCTIPlatform {
-  id: String!
+type OCTIPlatform implements Node {
```
jpkha, line 1: 💯

portal-api/src/modules/services/enrollment/enrollment.app.ts
jpkha: I think it's fine, and it's actually important to have some differences between the database model and the GraphQL model. However, we might need later for `OctiInstance` to inherit from `ServiceInstance` node. That would make things clearer on the frontend, especially for typing and if we ever want to filter between different `ServiceInstance` types.

portal-front/src/components/service/home/hooks/useServiceQueries.ts
jpkha: I think you are looking for `plural` directive : 
image
You shouldn't need `unknown`

portal-front/src/components/service/service-instance-card.tsx
jpkha: Can we make a specific component for this type of description ?
Also I think could be great to have a ReactNode here instead of ternary condition IMO

portal-front/src/components/service/home/owned-services.tsx
```diff
@@ -8,16 +10,44 @@ import ServiceInstanceCard from '../service-instance-card';
...
+      identifier: ServiceDefinitionIdentifierEnum.OCTI_ENROLLMENT,
+    },
+    links: [
+      {
+        url: instance.url,
+        name: instance.url,
+      },
+    ],
+    ordering: -1, // OCTI Instances are displayed at the first position
+  } as unknown as serviceList_fragment$data;
```
jpkha, line 35: I think this happens because the GraphQL schemas are different.
If you set up proper inheritance in GraphQL across the different services, you can avoid this issue and help TypeScript understand that it’s the same object.

jbanety: Sorry. This is a forgotten line.
With the refactor I can delete it.

portal-front/src/components/service/home/hooks/useServiceQueries.ts
jbanety: Nice 👍 
I give it a try

portal-api/src/modules/services/enrollment/enrollment.app.ts
jbanety: This is unrelated to DB.
I think we need to separate the concern here.

portal-api/src/modules/services/enrollment/enrollment.domain.ts
jbanety: The service must be displayed even if I personally do not have a UserService associated with the subscription of my organization.
I therefore cannot make a stronger security check than selecting in relation to the active organization.

portal-api/src/modules/services/enrollment/enrollment.domain.ts
jbanety: 🔝 
=> https://github.com/FiligranHQ/xtm-hub/pull/776#discussion_r2212745661

portal-api/src/modules/services/enrollment/enrollment.domain.ts
```diff
@@ -102,4 +110,42 @@ export const enrollmentDomain = {
       targetOrganizationId,
     });
   },
+
+  octiInstances: async (
```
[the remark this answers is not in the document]
jbanety: => https://github.com/FiligranHQ/xtm-hub/pull/776/commits/bb1d8c0ef5f1b056df3c25bce3d3e81f53a68713

portal-front/src/components/service/home/hooks/useServiceQueries.ts
jbanety: => https://github.com/FiligranHQ/xtm-hub/pull/776/commits/9ec456d3b81d5c4ea7084cd387ca7d04fb166d88

portal-api/src/modules/services/enrollment/enrollment.domain.ts
jbanety: I tried and this will lead to an unrecoverable error as the layer doesn't exist.
I can create an empty layer but I don't like this idea. WDYT?

portal-front/src/components/enroll/enroll.graphql.ts
```diff
@@ -31,3 +31,27 @@ export const EnrollOCTIInstance = graphql`
...
+    url
+    contract
+  }
+`;
+
+export const enrollOCTIInstanceListFragment = graphql`
+  fragment enrollOCTIInstanceListFragment on Query
+  @refetchable(queryName: "EnrollOCTIInstanceListQuery") {
+    octiInstances {
+      ...enrollOCTIInstanceFragment @relay(mask: false, plural: true)
```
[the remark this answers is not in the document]
jbanety: The directive @relay on fragment spreads for @inline fragments are not yet supported

portal-front/src/components/service/service.types.ts
```diff
@@ -0,0 +1,16 @@
...
+  publicServiceList_services$data['publicServiceInstances']['edges'][number]['node'];
+
+export enum JOIN_TYPE {
+  JOIN_INVITE = 'JOIN_INVITE',
+  JOIN_ASK = 'JOIN_ASK',
+  JOIN_AUTO = 'JOIN_AUTO',
+  JOIN_SELF = 'JOIN_SELF',
+}
+
+export enum SERVICE_CREATION_STATUS {
```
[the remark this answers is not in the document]
jbanety: => https://github.com/FiligranHQ/xtm-hub/pull/776/commits/456bd682896890ed199b55c2db1484d149a2d94b

apps/portal-front/test.Dockerfile
```diff
@@ -5,21 +5,30 @@ FROM node:23-alpine AS base
...
-# Install dependencies based on the preferred package manager
-COPY apps/portal-front/.yarnrc.yml apps/portal-front/package.json apps/portal-front/yarn.lock ./
+# Copy monorepo configuration files
+COPY .yarnrc.yml package.json yarn.lock ./
+COPY apps/portal-front/package.json ./apps/portal-front/package.json
+
+# Install all dependencies at the workspace level
 RUN corepack enable && \
-    yarn install --immutable
+    yarn install
```
jpkha, line 14: When running on CI, it's a good practice to keep --immutable

apps/portal-api/src/modules/services/enrollment/enrollment.app.ts
```diff
@@ -19,6 +20,20 @@ import {
 } from './enrollment.domain';
 
 export const enrollmentApp = {
+  loadOctiInstances: async (
+    context: PortalContext
+  ): Promise<OctiPlatform[]> => {
+    const instances = await enrollmentDomain.loadOctiInstances(context);
+    return instances.map((instance) => ({
+      __typename: 'OCTIPlatform',
```
jpkha, line 28: We `loadOctiInstances` shouldn't we call it `OCTIInstance` instead ?

apps/portal-api/Dockerfile
```diff
@@ -4,18 +4,24 @@ FROM node:23-alpine AS base
...
-# Install dependencies based on the preferred package managers
-COPY apps/portal-api/.yarnrc.yml apps/portal-api/package.json apps/portal-api/yarn.lock ./
+# Copy monorepo configuration files
+COPY .yarnrc.yml package.json yarn.lock ./
+COPY apps/portal-api/package.json ./apps/portal-api/package.json
+
+# Install all dependencies at the workspace level
 RUN corepack enable && \
-    yarn install --immutable
+    yarn install
```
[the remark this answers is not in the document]
jbanety: => https://github.com/FiligranHQ/xtm-hub/pull/776/commits/f1cc2106c1f045d88236a1eebebbbd7fa3cc808b

apps/portal-api/src/modules/services/enrollment/enrollment.app.ts
jbanety: => https://github.com/FiligranHQ/xtm-hub/pull/776/commits/96017e3ab4a2784d858285d9016f4e420840883b

portal-api/src/modules/services/enrollment/enrollment.domain.ts
hervyt: The layer will not be empty as we can check we send back only the instances of the organization (even if it's already checked here, it will be useful for every other select on this entity). See the JPK's presentation about the securityLayer where we talked about it 😉

jbanety: https://github.com/FiligranHQ/xtm-hub/pull/776/commits/745344c746d437b348104c656b2d2d9407aded69

Issue #470, opened 2025-04-01
For a user with manage access or admin orga capability, ability to accept other users from the same organization in his/her organization


EllynBsc:
**Overall Description**
For a user with manage access or admin orga capability, ability to accept other users from the same organization in his/her organization

PR #806, opened 2025-07-29
[backend] feat(pendingUsers/chunk1): add users to pending users table

carinelebas:
# Context: 
> Create pending users table and add the newly invited users in it if they were not invited by an admin.

# How to test:  
> Unit tests and local tests were done. 
To test it:
- we need to check the user is well added to the table when he is created from Auth0
- we need to check that when a user is added in user_organization, he is removed from user_organization_pending.

This is covered with unit tests.

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [x] Local tests

# Additional information:

Related #470


apps/portal-api/src/modules/common/user-organization-pending.domain.ts
```diff
@@ -0,0 +1,13 @@
+import UserOrganizationPending from '../../model/kanel/public/UserOrganizationPending';
+import { dbUnsecure } from '../../../knexfile';
+import { UserId } from '../../model/kanel/public/User';
+import { OrganizationId } from '../../model/kanel/public/Organization';
+
+export const removeUserFromOrganizationPendingUnsecure = async (
```
hervyt, line 22: **issue**: you should not use unsecure here but a secureQuery

apps/portal-api/src/modules/common/user-organization.domain.ts
```diff
@@ -23,6 +23,14 @@ export const insertNewUserOrganization = (
   return db(context, 'User_Organization').insert(field).returning('*');
 };
 
+export const insertNewUserOrganizationUnsecure = (
+  field: UserOrganizationInitializer | UserOrganizationInitializer[]
+): Promise<UserOrganization[]> => {
+  return dbUnsecure<UserOrganization>('User_Organization')
```
hervyt, line 29: issue: you should not use unsecure here but a secureQuery

apps/portal-api/src/modules/common/user-organization.helper.ts
```diff
@@ -0,0 +1,25 @@
+import { UserId } from '../../model/kanel/public/User';
+import { OrganizationId } from '../../model/kanel/public/Organization';
+import UserOrganization, { UserOrganizationInitializer } from '../../model/kanel/public/UserOrganization';
+import { insertNewUserOrganizationUnsecure } from './user-organization.domain';
+import { removeUserFromOrganizationPendingUnsecure } from './user-organization-pending.domain';
+
+
+export const createUserOrganizationRelationUnsecure = async ({
```
hervyt, line 8: **question** why is it called unsecure here ?

apps/portal-api/src/security/layer/user-organization-pending.ts
```diff
@@ -0,0 +1,32 @@
+import { KnexQueryBuilder } from '../../../knexfile';
+import { SecuryQueryHandlers } from '../access';
+
+const setDeleteSecurity = (): KnexQueryBuilder => {
+  // Implement User_organization_pending-specific insert security logic
+  throw new Error('Missing security logic');
```
hervyt, line 6: **question**: You don't need the capability to do that ? 
```
await checkUserCapabilities(context, [
    OrganizationCapability.AdministrateOrganization,
    OrganizationCapability.ManageAccess,
  ]);
  return qb;
```

apps/portal-api/src/modules/users/users.test.ts
```diff
@@ -18,18 +18,23 @@ import {
   preventAdministratorRemovalOfOneOrganization,
   removeUser,
 } from './users.helper';
+import { loadUserOrganizationPendingUnsecure } from '../common/user-organization-pending.domain';
+import { PLATFORM_ORGANIZATION_UUID } from '../../portal.const';
 
 describe('User helpers', async () => {
   describe('createNewUserFromInvitation', () => {
-    it('should create a new user with Role USER and not add in an existing Organization', async () => {
+    it('should create a new user with Role USER and not add in an existing Organization, but in pending organization', async () => {
```
hervyt, line 26: 💯

apps/portal-api/src/modules/common/user-organization.helper.ts
jbanety: Should be executed in a "system" context as it comes from the Auth0 callback before the user is created

apps/portal-api/src/modules/users/users.test.ts
```diff
@@ -62,6 +71,24 @@ describe('User helpers', async () => {
...
+      });
+      const newUser = await loadUserBy({ email: testMail });
+      const newUserPendingOrg = await  loadUserOrganizationPendingUnsecure({user_id: newUser.id});
+      expect(newUser).toBeTruthy();
+      expect(newUser.selected_org_capabilities.length).toBe(1);
+
+      expect(newUserPendingOrg.length).toBe(0);
+
+      // Delete corresponding in order to avoid issue with other tests
+      await removeUser(contextAdminUser, { email: newUser.email });
```
Kroustille, line 89: Is it necessary ? I think if the domain is not the same as the other ones in this tests, it should work.

We also reset data at every test file 🙂

apps/portal-api/src/modules/common/user-organization.helper.test.ts
```diff
@@ -0,0 +1,51 @@
...
+let user:UserLoadUserBy | null;
+
+describe('UserOrganizationHelper', () => {
+  describe('createUserOrganizationRelationUnsecure', () => {
+    afterEach(async () => {
+      if (user) {
+        await removeUser(contextAdminUser, { email: user.email });
+        user = null;
+      }
+    });
```
Kroustille, line 20: I think it's not necessary 🙂

apps/portal-api/src/modules/common/user-organization.domain.ts
carinelebas: I can't easily add it here: 
The call stack is starts from Auth0 callback [login-handle](https://github.com/FiligranHQ/xtm-hub/blob/d3a403b52bbc16285b94d4aa9022d7a7bcb85e8d/apps/portal-api/src/auth/login-handle.ts#L7), where I can't build the mandatory express request and response. 
Call stack is:
loginFromProvider > getOrCreateUser > createNewUserFromInvitation > createNewUserWithPendingOrga > insertNewUserOrganizationPendingUnsecure

getOrCreateUser is also called by the mutation addUserService. If we want to "secure" it one day, we'll have to duplicate the code or have a context that can fit calls from Auth0 ? 

Note that there is no regression here, the queries were already "unsecure".

apps/portal-api/src/modules/common/user-organization.helper.ts
carinelebas: I initially simply moved the function. Though I removed it now. We can know if a query is secure or not depending is the context is passed.

apps/portal-api/src/security/layer/user-organization-pending.ts
carinelebas: Indeed, I added it. Since it was not used, it was not an issue.

apps/portal-api/src/modules/users/users.test.ts
carinelebas: I removed it 👍

apps/portal-api/src/modules/common/user-organization.helper.test.ts
carinelebas: I removed it 👍

apps/portal-api/src/modules/common/user-organization.domain.ts
jbanety: Here we are talking about `insertNewUserOrganizationUnsecure` not `insertNewUserOrganizationPendingUnsecure`.
Inserting a new user <> organization relation must be secured.

carinelebas: Sorry I made a typo, though actually, it can't be easily secured either: 
Functionally speaking: when we create a user from Auth0, we don't have a context, but we need to add the relation between a user and its personal organization, so we can't secure the insert in this case. 

Technically, it means when coming from `createNewUserFromInvitation` , or `createUserWithPersonalSpace`, which are also used by several other functions. 
We would need to do some refactoring on the users creation to isolate this flow, or have a context when coming from Auth0 to make it secure, but I think this should rather be done in a technical PR ? 

I extracted this piece of code into a function, but it was originally in [createUserOrganizationRelationUnsecure](https://github.com/FiligranHQ/xtm-hub/blob/d3a403b52bbc16285b94d4aa9022d7a7bcb85e8d/apps/portal-api/src/modules/common/user-organization.domain.ts#L52C14-L52C52), so it was already not secure.

jbanety (approved):

Kroustille (approved):

apps/portal-api/src/modules/common/user-organization.helper.ts
```diff
@@ -0,0 +1,40 @@
...
+  organizations_id.map(org => removeUserFromOrganizationPending(context, user_id, org));
+
+  return createUserOrganizationRelation({user_id, organizations_id})
+};
+
+
+export const createUserOrganizationRelation = async ({
+                                                                       user_id,
+                                                                       organizations_id = [],
+                                                                     }: {
```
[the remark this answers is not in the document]
carinelebas: Good point, thx! I pushed the linted files!

Issue #519, opened 2025-04-15
Save our subscription’s dates in date_with_timezones


EllynBsc:
**Overall Description**
At the moment, we save our dates in dates format (without timezones, format AAAA-MM-DD)

We need to save our subscription’s dates in dates with timezones (just like the `last_login `column in user, that is type `datetime`) to take into account the different timezones.
> Create a new migration to change to column `start_date `and `end_date ` type)

PR #829, opened 2025-08-05
[backend] - fix(dates) Change dates into dateTime #519

hervyt:
# Context: 
> IN the DB types, change the time into dateTime to have a full date model 

# How to test:  
There should be no regression 

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [X] Local tests

# Additional information:

Related #519


jpkha (approved):

apps/portal-api/src/migrations/20250805083254_pass_date_into_datetime.js
```diff
@@ -0,0 +1,89 @@
...
+  await knex.schema.alterTable('Subscription', (table) => {
+    table.dateTime('end_date').nullable();
+  });
+
+  await knex.schema.alterTable('Service_Price', (table) => {
+    table.dropColumn('start_date');
+  });
+  await knex.schema.alterTable('Service_Price', (table) => {
+    table.dateTime('start_date').nullable();
+  });
```
Kroustille, line 43: I tried using raw SQL and you can directly change the column type to timestamp (I think equivalent to knex dateTime) by doing

```
ALTER TABLE "Subscription"
ALTER COLUMN start_date
TYPE timestamp;
```

So I think you can use:
```
await knex.schema.alterTable('Subscription', function (table) {
  table.dateTime('start_date').alter();
});
```

Kroustille (approved):
I think we should launch `generate-pg-to-ts` script from `portal-api` either in a precommit hook or when the dev server is launched.

`kanel` was updated in June but I'm thinking nobody launched `generate-pg-to-ts` since then and that's why you have a bunch of kanel updated 😉

jpkha (approved):

Issue #579, opened 2025-05-06
[Tech Debt/backend] - [JoiningService] : Handle error in the backend for ERROR_SERVICE_INSTANCE_USER_MUST_JOIN_SERVICE


hervyt:
**Overall Description**
On loading detailed page of resources sharing, we handle an error from the backend : 
ERROR_SERVICE_INSTANCE_USER_MUST_JOIN_SERVICE
We should handle the error in the backend but not in the frontend 

**Additional context**

![Image](https://github.com/user-attachments/assets/f2a5e927-e575-4110-b535-6a9f28d5e22c)

**Solution**
Use a "joinService" mutation instead.

PR #836, opened 2025-08-06
[backend/frontend]: improve way to get service and add OBAS Scenario email #579 #843

hervyt:
# Context: 
> Code refacto on the way to access service 
> Send email on the first access to the OBAS scenario 

# How to test:  
Subscribe to the OBAS Scenario librairy as user1
The subscriber (user1) should receive an email for OBAS Scenario lib. 
Connect as user2, in the same organization as user1. 
You should see the tile "OBAS Scenario lib" without any subscription needed. Click on the tile **for the first time**. user2 should receive an email for OBAS Scenario lib. 

You should see no regression elsewhere, specially on : 
- Custom Dashboard lib 
- CSV lib
- subscription administration as BYPASS 
- "MANAGE ACCESS" button workflow (as user with the capa "MANAGE_ACCESS" on a lib and that is not BYPASS. 

# What tests has been made: 
- [X] Integration tests
- [ ] E2E tests
- [X] Local tests

# Additional information:

Related #579 #843

### DIFF
--- a/apps/portal-api/src/modules/services/definition/service-capability/service-capability.domain.test.ts
+++ b/apps/portal-api/src/modules/services/definition/service-capability/service-capability.domain.test.ts
@@ -1,0 +1,145 @@
+import { v4 as uuidv4 } from 'uuid';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { dbUnsecure } from '../../../../../knexfile';
+import { contextAdminUser } from '../../../../../tests/tests.const';
+import ServiceCapability, {
+  ServiceCapabilityId,
+} from '../../../../model/kanel/public/ServiceCapability';
+import ServiceDefinition, {
+  ServiceDefinitionId,
+} from '../../../../model/kanel/public/ServiceDefinition';
+import { loadServiceCapabilitiesBy } from './service-capability.domain';
+
+describe('Service Capability domain', () => {
+  let testServiceDefinitionId: ServiceDefinitionId;
+  let testServiceDefinitionId2: ServiceDefinitionId;
+  let testCapabilityIds: ServiceCapabilityId[] = [];
+  let testServiceDefIds: ServiceDefinitionId[] = [];
+
+  beforeAll(async () => {
+    const serviceDefinition = await dbUnsecure<ServiceDefinition>(
+      'ServiceDefinition'
+    )
+      .insert([
+        {
+          id: uuidv4() as ServiceDefinitionId,
+          name: 'Test Service Definition',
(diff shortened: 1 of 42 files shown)


apps/portal-api/src/modules/services/definition/service-capability/service-capability.domain.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        .whereIn('id', testCapabilityIds)
+        .delete();
+    }
+
+    if (testServiceDefinitionId) {
+      await dbUnsecure<ServiceDefinition>('ServiceDefinition')
+        .whereIn('id', testServiceDefIds)
+        .delete();
+    }
+  });
```
Kroustille, line 83: I think this is not necessary and we could use a `beforeAll` instead of a `beforeEach`

apps/portal-api/src/modules/services/definition/service-capability/service-capability.domain.ts
```diff
@@ -0,0 +1,21 @@
...
+    | addPrefixToObject<ServiceCapabilityMutator, 'Service_Capability.'>
+    | ServiceCapabilityMutator
+): Promise<ServiceCapability[]> => {
+  const servicesCapa: ServiceCapability[] = await db<ServiceCapability>(
+    context,
+    'Service_Capability'
+  )
+    .where(field)
+    .select('Service_Capability.*');
+  return servicesCapa;
```
Kroustille, line 20:
```suggestion
  return db<ServiceCapability>(
    context,
    'Service_Capability'
  )
    .where(field)
    .select('Service_Capability.*');
```

or

```suggestion
  const serviceCapabilities: ServiceCapability[] = await db<ServiceCapability>(
    context,
    'Service_Capability'
  )
    .where(field)
    .select('Service_Capability.*');
  return servicesCapabilities;
```

apps/portal-api/src/modules/services/definition/service-definition.resolver.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      ).rejects.toThrow('Failed to load capabilities');
+    });
+  });
+});
+
+afterEach(async () => {
+  vi.clearAllTimers();
+  vi.restoreAllMocks();
+  vi.clearAllMocks();
+});
```
Kroustille, line 113: This block should be in the first `describe` block. I always recommend to put it at the beginning of the block because it's easier to know it exists before reading tests.

Kroustille (approved):
It's great to see removed dead code and new tests, congrats 👏

PR #839, opened 2025-08-07
[frontend/backend] - feature(pendingUsers/chunk2) Display list of pending users and allow accept/reject #470

carinelebas:
# Context: 
> We should display a new table containing users that belong to the same domain as the users but not yet part of the organization. 

In a previous PR, users having a mail in the same domain were added to the new table `user_organization_pending`.
In this PR:
- On users page, add user button is moved at title level
- we display a new tab on the users page, containing the users in this table (not on the admin part)
- we can accept the users (move them to the `user_organization` table) or reject them (remove them from `user_organization_pending`)
- users lists and pending users list displayed are updated live thanks to graphql subscriptions
- pagination can accept an 'unsecure' option to run the query without side effects. 

Edit:
- Add fix for admin users that are added at first log in in pending_user_organization and user_organization_table
- add migration to remove from user_organization_pending the users that may already be in this case

# How to test:  
- Check the pending users are well displayed
- Check they can be accepted, they disappear from the pending user list, and appear in the users list
- Check they can be rejected, they disappear forever from the pending user list
- Check the admin/security still works as expected.

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [x] Local tests

# Additional information:

Related #470


apps/portal-api/src/modules/users/users.domain.ts
```diff
@@ -350,6 +350,45 @@ export const loadUsers = (context: PortalContext, opts: QueryUsersArgs) => {
...
+
+  return paginate<UserGenerated, UserConnection>(
+    context,
+    'User',
+    {
+      ...opts,
+      filters: filters?.map(({ key, value }) => {
+        if (key === FilterKey.OrganizationId) {
+          return {
+            key: 'UserOrgPendingFilter.organization_id',
```
jpkha, line 380: Maybe a dumb question but do you know why do we need to return `UserOrgPendingFilter.organization_id` instead of `organization_id` ?
The 'as unknown as Filter' below is bothering me 👯

apps/portal-api/src/modules/users/users.graphql
```diff
@@ -132,7 +142,12 @@ type MeUserSubscription {
...
 
+type UserPendingSubscription {
+  delete: User
+}
+
 type Subscription {
   User: UserSubscription @auth
   MeUser: MeUserSubscription @auth
+  UserPending: UserPendingSubscription @auth
 }
```
jpkha, line 153:
```suggestion
type Subscription {
  User: UserSubscription @auth(requires: [ADMINISTRATE_ORGANIZATION, MANAGE_ACCESS])
  MeUser: MeUserSubscription @auth
  UserPending: UserPendingSubscription @auth(requires: [ADMINISTRATE_ORGANIZATION, MANAGE_ACCESS])
}
```
I think we can set up more restrictive capabilities as we only use in a organization level panel.

apps/portal-api/src/modules/users/users.resolver.ts
```diff
@@ -253,15 +271,31 @@ const resolvers: Resolvers = {
...
+          organization_id,
           input.capabilities
         );
 
+        const [userOrganization] = await loadUserOrganization(context, {
+          user_id: userId,
+          organization_id,
+        });
+
+        if (!userOrganization) {
```
jpkha, line 288: I'm not completely confident in this implementation. I would prefer having a separate endpoint for the pending user use case. Alternatively, we could define two functions in the domain/app layer and call the appropriate one based on this condition.
WDYT ?

apps/portal-api/src/security/user-security-access.ts
```diff
@@ -21,5 +21,15 @@ export const meUserSSESecurity = (opt: {
...
+    isUserGranted(opt.user, OrganizationCapability.ManageAccess) ||
+    isUserGranted(opt.user, OrganizationCapability.AdministrateOrganization)
+  );
+};
+
+export const userPendingSSESecurity = (opt: { user: UserLoadUserBy }) => {
+  return (
+    isUserGranted(opt.user, OrganizationCapability.ManageAccess) ||
+    isUserGranted(opt.user, OrganizationCapability.AdministrateOrganization)
+  );
```
jpkha, line 34: 💯

apps/portal-front/app/(application)/app/(user)/manage/user/page-loader.tsx
```diff
@@ -20,14 +19,12 @@ const breadcrumbValue = [
...
 const PageLoader: React.FunctionComponent = () => {
-  const t = useTranslations();
   const { me } = useContext(PortalContext);
 
   return (
     <>
       <BreadcrumbNav value={breadcrumbValue} />
-      <h1 className="sr-only">{t('UserListPage.Title')}</h1>
-      <UserList organization={me?.selected_organization_id} />
+      <UserListPage organization={me?.selected_organization_id}></UserListPage>
```
jpkha, line 27:
```suggestion
      <UserListPage organization={me?.selected_organization_id}/>
```

apps/portal-front/src/components/admin/user/user.graphql.ts
```diff
@@ -10,8 +10,16 @@ export const UserListCreateMutation = graphql`
...
 export const UserSlugEditMutation = graphql`
-  mutation userSlugEditMutation($id: ID!, $input: EditUserCapabilitiesInput!) {
-    editUserCapabilities(id: $id, input: $input) {
+  mutation userSlugEditMutation(
+    $id: ID!
+    $input: EditUserCapabilitiesInput!
+    $userListConnections: [ID!]!
+  ) {
+    editUserCapabilities(id: $id, input: $input)
+      @prependNode(
```
jpkha, line 19: Do you still need this prepend ? As you update the userList from the subscription, IMO you don't need this part.

apps/portal-front/src/components/admin/user/user-list.tsx
```diff
@@ -308,7 +329,7 @@ const UserList: FunctionComponent<UserListProps> = ({ organization }) => {
   );
 
   return (
-    <UserListContext.Provider value={{ connectionID: data.users.__id }}>
+    <div>
```
jpkha, line 332:
```suggestion
    <>
```

Kroustille (approved):

apps/portal-api/src/modules/users/users.graphql
carinelebas: done in [e07b74f](https://github.com/FiligranHQ/xtm-hub/pull/839/commits/e07b74fc4db20b798eac5af19c844eef316df1d7)

jpkha:
You should rebase instead of merge in the feature branch.
image
By rebasing you can avoid to have a merge commit.

Issue #849, opened 2025-08-12
Ability for user to register OpenCTI platform from local environment to Hub


EllynBsc:
**Overall Description**

At the moment if you have an OpenCTI platform running locally, you cannot register your platform in the Hub due to security layers.

We need to allow users to do so and add documentation here: https://docs.hub.filigran.io/latest/user/opencti-registration/

**Solution**

- add a new mutation on the hub to create or refresh user token
- check the user token in download document query
- when OpenCTI tab is opened
 - if it has token in session storage
   - it uses it to fetch the desired resource
   - if the request failed because of authentication, OpenCTI do the token refresh flow below
 - else 
   - OpenCTI tells the hub it's ready to receive a token via postMessage
   - when Hub receives the postMessage, it calls the new mutation to have a user token
   - hub responds to OpenCTI via postMessage with the token
   - OpenCTI saves it in sessionStorage
   - OpenCTI uses the token to fetch the desired resource

jbanety:
The reason why we can’t simply use the existing Hub cookies from our app running on localhost is due to browser security restrictions (`Same-Origin Policy`, third-party cookie blocking). Even with `SameSite=None` and `Secure`, most modern browsers will block sending cookies cross-origin from localhost to hub.filigran.io.

Proposed solution:
Instead of relying on the Hub session cookie, we should use the token provided when the Platform is registered by the Hub. This token could then be sent from the Platform to a dedicated REST or GraphQL endpoint on the Hub.
That endpoint would validate the token and return the requested resource.

This approach avoids browser cookie restrictions and provides a clear, token-based authentication flow between the Platform and the Hub.

PR #859, opened 2025-08-14
[backend/frontend] feat: add ability to download document with new token

Kroustille:
# Context: 

- add platform token to users
- provide mutation to refresh platform token
- communicate with OpenCTI via postMessage to send token

# How to test:  

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [x] Local tests

# Additional information:

Related #849

### DIFF
--- a/apps/portal-api/src/modules/services/document/document-download-endpoint.ts
+++ b/apps/portal-api/src/modules/services/document/document-download-endpoint.ts
@@ -1,19 +1,19 @@
 import cors from 'cors';
+import { Request } from 'express';
 import rateLimit from 'express-rate-limit';
 import { fromGlobalId } from 'graphql-relay/node/node.js';
 import { Readable } from 'stream';
-import { dbTx, dbUnsecure } from '../../../../knexfile';
-import Document, { DocumentId } from '../../../model/kanel/public/Document';
+import { dbTx } from '../../../../knexfile';
+import { DocumentId } from '../../../model/kanel/public/Document';
+import { ServiceInstanceId } from '../../../model/kanel/public/ServiceInstance';
+import { PortalContext } from '../../../model/portal-context';
+import { UserLoadUserBy } from '../../../model/user';
 import { logApp } from '../../../utils/app-logger.util';
 import { NotFoundError } from '../../../utils/error.util';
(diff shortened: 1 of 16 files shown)


apps/portal-api/src/modules/services/document/document-download-endpoint.ts
```diff
@@ -27,18 +32,35 @@ export const documentDownloadEndpoint = (app) => {
...
+        if (!platform_token) {
+          res.status(401).json({ message: 'You must be logged in' });
+          return;
+        }
+        user = await loadUserBy({ 'User.platform_token': platform_token });
+        if (!user) {
+          res.status(401).json({ message: 'You must be logged in' });
+          return;
+        }
+      }
```
jpkha, line 47: We have two time error 401 with !User, can we avoid it?

apps/portal-api/src/modules/services/document/document-download-endpoint.ts
```diff
@@ -27,18 +32,35 @@ export const documentDownloadEndpoint = (app) => {
...
-        const [document] = await dbUnsecure<Document>('Document')
-          .where(
-            'Document.id',
-            '=',
-            fromGlobalId(req.params.filename).id as DocumentId
-          )
-          .select('Document.*');
+        const context: PortalContext = {
+          user: user,
+          serviceInstanceId: fromGlobalId(req.params.serviceInstanceId)
```
jpkha, line 55: You can use extractId instead of fromGlobalId().id

apps/portal-front/src/components/service/document/one-click-deploy/one-click-deploy.tsx
```diff
@@ -30,38 +31,42 @@ export const OneClickDeployOctiPlatformsQuery = graphql`
...
   };
 
+  useEffect(() => {
+    if (!shouldOpenTab) {
+      return;
+    }
+
+    openTab();
+    setShouldOpenTab(false);
+  }, [shouldOpenTab]);
```
jpkha, line 68: Are you sure you need a UseEffect here? The useState of shouldOpenTab should be enough to trigger a rerender and trigger openTab() IMO

Issue #882, opened 2025-08-28
One-click deploy works even if the connectivity is lost


EllynBsc:
**Describe the bug**
Current: And I notice that the one-click deploy works even if the connectivity is lost (or appears as such in the front end).

Expected: the one-click deploy button on the bug should not work if the connectivity is lost and should display a message on OCTI platform:

If user does not have the capability to register:
```
Modal title: Connectivity Lost
Modal text: The connection between your OpenCTI platform and the XTM Hub is currently unavailable. Some features such as the one-click deploy may not function as expected until connectivity is restored.
Please contact your OpenCTI administrator to resolve the issue.
```

If user has the capability to register:
```
Modal title: Connectivity Lost
Modal text: The connection between your OpenCTI platform and the XTM Hub is currently unavailable. Some features such as the one-click deploy may not function as expected until connectivity is restored.
Please re-register your OpenCTI platform to enable the one-click deployment.
```

EllynBsc:
@Kroustille, good question:
Let's have a 'cancel' CTA indeed that redirect to the library he/she was on ( custom dashboard, integration feeds, scenarios later)

jpkha:
Release in v1.7.4

PR #888, opened 2025-08-29
[backend] feat: assert platform token and id to download resource

Kroustille:
# Context: 

- check platform token and platform id to allow user to download document
- check only when user platform token is provided (one click deploy case)

# How to test:  

- try to download document with cookies but with user platform token
- check that you're not allowed
- try to download with invalid platform id and token couple
- check that you're not allowed
- try to download with valid platform id and token couple

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [x] Local tests

# Additional information:

Related #882

### DIFF
--- a/apps/portal-api/src/modules/services/document/document-download-endpoint.ts
+++ b/apps/portal-api/src/modules/services/document/document-download-endpoint.ts
@@ -18,6 +18,7 @@ import {
   shouldSendEventForService,
 } from '../../telemetry/telemetry.helper';
 import { loadUserBy } from '../../users/users.domain';
+import { serviceContractDomain } from '../contract/domain';
 import { loadServiceDefinitionByServiceInstance } from '../service-instance.domain';
 import { downloadFile } from './document-storage';
 import { incrementDocumentsDownloads, loadDocumentBy } from './document.domain';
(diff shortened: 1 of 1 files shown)


apps/portal-api/src/modules/services/document/document-download-endpoint.ts
```diff
@@ -22,17 +23,29 @@ const documentDownloadRateLimiter = rateLimit({
...
+    return { user: null };
   }
 
-  return loadUserBy({ 'User.platform_token': platform_token });
+  const userLoadFromUserPlatformToken = await loadUserBy({
+    'User.platform_token': user_platform_token,
+  });
+  return {
+    user: userLoadFromUserPlatformToken,
+    isLoadedFromUserPlatformToken: true,
```
jpkha, line 47: Pourquoi tu as besoin d'un boolean ? ça serait mieux d'avoir une fonction à part pour éviter de complexifier le load user ?

apps/portal-api/src/modules/services/document/document-download-endpoint.ts
```diff
@@ -44,25 +57,51 @@ export const documentDownloadEndpoint = (app) => {
...
+          ),
+          req,
+          res,
+        };
+
+        if (isLoadedFromUserPlatformToken) {
+          const token = req.header('XTM-Hub-Platform-Token');
+          if (!token) {
+            res.status(403).json({ message: 'missing token in headers' });
+            return;
```
jpkha, line 79:
```suggestion
            return res.status(403).json({ message: 'missing token in headers' });
```
Should be the same ?

jpkha (approved):

Issue #918, opened 2025-09-04
[BUG] : User in personal space shouldn't not access or manage users


jpkha:
**Describe the bug**
As user, we should not have the capability to access to the user list in personal space

**Screenshots and screen records**

<img width="1917" height="447" alt="Image" src="https://github.com/user-attachments/assets/13d0c091-edac-40a4-8765-3ee9c9ee378d" />

**To Reproduce**

- Connect on your personal space
- Access to user list

PR #945, opened 2025-09-10
[frontend] - (USersList): No users on personal space #918

hervyt:
# Context: 
> Connected, you should not see the users list when you are on your personal space. 

# How to test:  
Connect as Admin_orga 
Switch to personal space 
You should not see the users list 
Connect as User without any special capabilities
Switch to personal space 
You should not see the users list 

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [X] Local tests

# Additional information:

Related #918

### DIFF
--- a/apps/portal-front/src/components/navigation.tsx
+++ b/apps/portal-front/src/components/navigation.tsx
@@ -17,7 +17,11 @@ export const NavigationApp: FunctionComponent<NavigationAppProps> = ({
   open,
 }) => {
   const t = useTranslations();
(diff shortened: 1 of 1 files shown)


carinelebas (approved):
All good! 
Though I have a question: If we are on the personal space, we don't see the users entry in the menu. But if we enter in the url `/app/manage/user`, we see can anyway access the users screen. Is it a voluntary backdoor, or should we also restrict the access here ? 😄

carinelebas (approved):

PR #950, opened 2025-09-11
[backend/frontend] display openAEV services #949

carinelebas:
# Context: 
> We need to display openAEV registered platforms. Testing the feature flag should not be needed since creation should already be protected by feature flag. 

# How to test:  
- Register an openAEV platform. Check it's well displayed in the front (note the ugly openBas header in the image will be changed later)
- Check an OpenCTI platform is still well displayed.

# What tests has been made: 
- [x] Integration tests
- [x] E2E tests
- [x] Local tests

# Additional information:

Related #949


apps/portal-front/public/openaev_registration-private-platform-logo.png
jpkha, line 1: Isn't it weird to have a file name openaev but display OpenBAS ?

carinelebas: done in [16b978d](https://github.com/FiligranHQ/xtm-hub/pull/950/commits/16b978de04279d05e1648a398a40a1c40df527cd)

PR #951, opened 2025-09-12
[backend] chore: add automatic GraphQL error mapper

Kroustille:
# Context: 

- add new enums to classify errors
- add GraphQL error mapper
  - map error to GraphQL error depending on enum
  - return unknown error when error is not mapped
  - use customer unknown error when specified

# How to test:  

- throw a mapped error in a resolver or app or domain
  - error is mapped to right GraphQLError
- throw an unmapped error in a resolver or app or domain
  - unknown error is thrown
- throw an unmapped error in a resolver or app or domain but with custom error in resolver
  - unknown error is thrown

# What tests has been made: 
- [ ] Integration tests
- [ ] E2E tests
- [x] Unit tests
- [x] Local tests

### DIFF
--- a/apps/portal-api/src/sessionStoreManager.ts
+++ b/apps/portal-api/src/sessionStoreManager.ts
@@ -1,7 +1,7 @@
 import { MemoryStore } from 'express-session';
 import { UserLoadUserBy, UserWithOrganizationsAndRole } from './model/user';
 import { logApp } from './utils/app-logger.util';
-import { UnknownError } from './utils/error/error.util';
+import { UnknownErrorCode } from './utils/error/error.code';
 
 let memoryStore = undefined;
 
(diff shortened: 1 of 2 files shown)


carinelebas (approved):

apps/portal-api/src/modules/users/users.resolver.ts
```diff
@@ -137,7 +137,8 @@ const resolvers: Resolvers = {
         );
 
         if (chosenOrganization.personal_space) {
-          throw new Error('CANT_ADD_USER_TO_PERSONAL_SPACE');
+          logApp.warn('You cannot add a user in your personal space');
```
jpkha, line 140: Why are we logging a warning here? Shouldn’t it be an error instead? Although, if we’re throwing an error, wouldn’t that already get logged?

PR #971, opened 2025-09-18
Chore: Update MinIO Version to RELEASE.2025-06-13T11-33-47Z

troll-os:
# Context
The current MinIO version was more than a year old by now and to align with other products, we're upgrading the version here. 

A run test was made by JPK on its dev env without issue. 

We'll deploy this version of the chart on xtmhub dev and staging following the 24h principle.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: xtmhub
 description: A Helm chart for XTM-Hub
 type: application
-version: 0.1.3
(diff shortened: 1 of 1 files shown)


nicogigi92 (commented):
yes we can also align here the version, but should we manage this version on the platform side, to centralise the management of renovations and upgrades there?
Furthermore, do we need to change the version of the chart used by our playbooks on the platform side?

nicogigi92 (approved):

PR #1102, opened 2025-10-15
[backend] feature(freetrials/chunk3): create trial request #1095

carinelebas:
# Context

We need to have a graphql endpoint that will create a freetrial request.

on Apollo sandbox, call the following mutation and check the result. 
<details>

```

mutation CreateDeploymentRequest {
  createDeploymentRequest(
    input: {
      platform_identifier: opencti
      region: europe
      type: trial
      intention: "Evaluating OpenCTI for threat intelligence management in our security operations"
      job_title: "Security Analyst"
      activity_sector: "Financial Services"
      use_case: "use case"
    }
  ) {
    id
    platform_identifier
    region
    type
    intention
    job_title
    activity_sector
    start_date
    end_date
    status
  }
}

```

</details>

## Related issues

- Related to #1095

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

### DIFF
--- a/apps/portal-api/src/modules/services/deployments/deployments.app.ts
+++ b/apps/portal-api/src/modules/services/deployments/deployments.app.ts
@@ -38,7 +38,7 @@ export const DeploymentsApp = {
         platformIdentifier: input.platform_identifier,
       });
 
-      const [createdDeploymentRequest] = await insertDeploymentRequest({
+      const createdDeploymentRequest = await insertDeploymentRequest({
         id: uuidv4() as DeploymentRequestId,
         user_requester_id: context.user.id,
         organization_requester_id: context.user.selected_organization_id,
--- a/apps/portal-api/src/modules/services/deployments/deployments.domain.ts
+++ b/apps/portal-api/src/modules/services/deployments/deployments.domain.ts
@@ -7,9 +7,12 @@ import DeploymentRequest, {
 export const insertDeploymentRequest = async (
   deploymentRequest: DeploymentRequestInitializer
 ) => {
-  return dbUnsecure<DeploymentRequest>('DeploymentRequest')
+  const [createdDeploymentRequest] = await dbUnsecure<DeploymentRequest>(
+    'DeploymentRequest'
+  )
     .insert(deploymentRequest)
     .returning('*');
+  return createdDeploymentRequest;
 };
 
 export const loadDeploymentRequestBy = async (
--- a/apps/portal-api/src/modules/services/instances/domain.ts
+++ b/apps/portal-api/src/modules/services/instances/domain.ts
@@ -16,7 +16,7 @@ export const serviceInstanceDomain = {
     context: PortalContext,
     serviceDefinitionId: string,
     platformIdentifier: PlatformIdentifier,
-    status: string = ServiceInstanceCreationStatus.Ready
+    creation_status: ServiceInstanceCreationStatus = ServiceInstanceCreationStatus.Ready
   ): Promise<ServiceInstanceId> => {
     const id = uuidv4() as ServiceInstanceId;
     await db(context, 'ServiceInstance').insert([
@@ -24,7 +24,7 @@ export const serviceInstanceDomain = {
         id,
         name: serviceInstanceNameMappedByPlatformIdentifier[platformIdentifier],
         description: '',
-        creation_status: status,
+        creation_status,
         public: false,
         join_type: 'JOIN_AUTO',
         tags: [
--- a/apps/portal-api/src/modules/services/deployments/deployments.app.test.ts
+++ b/apps/portal-api/src/modules/services/deployments/deployments.app.test.ts
@@ -29,21 +29,21 @@ describe('Deployment app', () => {
       const dbDeploymentRequest = await loadDeploymentRequestBy({
         id: deployment.id as DeploymentRequestId,
       });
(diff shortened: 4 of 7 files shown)


apps/portal-api/src/modules/services/deployments/deployments.app.ts
```diff
@@ -1,24 +1,77 @@
...
+      const [createdDeploymentRequest] = await insertDeploymentRequest({
+        id: uuidv4() as DeploymentRequestId,
+        user_requester_id: context.user.id,
+        organization_requester_id: context.user.selected_organization_id,
+        service_instance_id: serviceInstanceId,
+        status: DeploymentRequestStatus.Pending,
+        type: input.type,
+        platform_identifier: input.platform_identifier,
+        region: input.region,
+        intention: input.intention,
```
hervyt: **thought**: idk what this is...

apps/portal-api/src/modules/services/deployments/deployments.app.ts
```diff
@@ -1,24 +1,77 @@
...
+        job_title: input.job_title,
+        use_case: input.use_case,
+        activity_sector: input.activity_sector,
+        platform_token: uuidv4(),
+      });
+      await trx.commit();
+      return {
+        id: createdDeploymentRequest.id,
```
hervyt: **question**: Qhy can't you directly return createdDeploymentRequest and have to rebuild the object ?

apps/portal-api/knexfile.ts
```diff
@@ -148,8 +150,9 @@ export const db = <T>(
     opts
   );
 
-  if (context?.trx) {
-    queryContext.transacting(context.trx);
+  const reqContext = requestContext.get();
+  if (reqContext?.trx && !reqContext.trx.isCompleted()) {
+    queryContext.transacting(reqContext.trx);
```
Kroustille: Maybe we can add an helper function which does this because I think we'll need it in a lot of different places, and I'm afraid we're trying to use `transaction` on a completed transaction sometimes 😬

apps/portal-api/src/modules/services/instances/domain.ts
```diff
@@ -12,15 +15,16 @@ export const serviceInstanceDomain = {
...
+    status: string = ServiceInstanceCreationStatus.Ready
   ): Promise<ServiceInstanceId> => {
     const id = uuidv4() as ServiceInstanceId;
     await db(context, 'ServiceInstance').insert([
       {
         id,
         name: serviceInstanceNameMappedByPlatformIdentifier[platformIdentifier],
         description: '',
-        creation_status: 'READY',
+        creation_status: status,
```
Kroustille: If you apply the suggestion with the function parameters:
```suggestion
        creation_status,
```

apps/portal-api/src/modules/services/deployments/deployments.domain.ts
```diff
@@ -0,0 +1,22 @@
...
+  DeploymentRequestMutator,
+} from '../../../model/kanel/public/DeploymentRequest';
+
+export const insertDeploymentRequest = async (
+  deploymentRequest: DeploymentRequestInitializer
+) => {
+  return dbUnsecure<DeploymentRequest>('DeploymentRequest')
+    .insert(deploymentRequest)
+    .returning('*');
+};
```
Kroustille: This way you don't have to destructure the returned array in the caller. 
I personnally find it cleaner but it's up to you if you want to do the change 😉 
```suggestion
export const insertDeploymentRequest = async (
  deploymentRequest: DeploymentRequestInitializer
) => {
  const [createdDeploymentRequest] = await dbUnsecure<DeploymentRequest>('DeploymentRequest')
    .insert(deploymentRequest)
    .returning('*');
  return createdDeploymentRequest
};
```

apps/portal-api/knexfile.ts
carinelebas: This code is currently only used for register and unregister (and now deployment), which are the only places transaction is set in the context. 
Without the check `trx.isCompleted()`, it will throw an error if we try to execute queries on a completed transaction (I had the issues in tests). With it, it should be safe, since query will be executed outside the transaction.
We need a better way to manage the transactions, though to better manage the closed transaction. But I'd rather tackle it in tech pains. 
Wdyt?

apps/portal-api/src/modules/services/deployments/deployments.app.ts
carinelebas: They are not the same objects, and don't have the same fields: one is the kanel model (with fields to support platform data), the other one is the graphql object (only what is needed by the front). I would like to have an easier solution, but like this the conversion is clear at least.

apps/portal-api/src/modules/services/instances/domain.ts
carinelebas: done in [f19ea09](https://github.com/FiligranHQ/xtm-hub/pull/1102/commits/f19ea09d05939af65f1ad757710dbc51dca16a14)

apps/portal-api/knexfile.ts
Kroustille: Yes I agree it's better to tackle it in tech pain and it can be a broader subject 👍

hervyt (approved):

PR #1211, opened 2025-11-03
[backend] fix: allow unauthenticated user to access public data

Kroustille:
# Context

Integration feeds page is not accessible because `IntegrationFeed.service_instance` is protected by auth directive but needs to be accessed on the public page.

This PR removes the `@auth` directives to allow public rendering.

## Related issues

- Related to #1105 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

### DIFF
--- a/apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx
+++ b/apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx
@@ -18,14 +18,16 @@ const SlugDocument = async ({
   switch (serviceInstance.slug as ServiceSlug) {
     case ServiceSlug.OPEN_CTI_INTEGRATION_FEEDS:
       return (
-        <div className="relative w-full h-[35vh]">
-          <Image
-            fill
-            className="object-cover object-top rounded-lg"
-            src={`/document/images/${serviceInstance.id}/${document.children_documents![0]!.id}`}
-            alt={`A picture of ${document.name}`}
-          />
-        </div>
+        document.children_documents![0] && (
+          <div className="relative w-full h-[35vh]">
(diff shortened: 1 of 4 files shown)


apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx
```diff
@@ -18,14 +18,16 @@ const SlugDocument = async ({
...
-            src={`/document/images/${serviceInstance.id}/${document.children_documents![0]!.id}`}
-            alt={`A picture of ${document.name}`}
-          />
-        </div>
+        document.children_documents![0] && (
+          <div className="relative w-full h-[35vh]">
+            <Image
+              fill
+              className="object-cover object-top rounded-lg"
+              src={`/document/images/${serviceInstance.id}/${document.children_documents![0]!.id}`}
```
Comment (jpkha) [inline apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx]: Actually on creating an integration feeds can we create without image ? 
If we put "!" we can use a directive required https://relay.dev/docs/guides/required-directive/

But if we expect to be null sometime, we should manage this case

Comment (Kroustille) [reply apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx]: At least I had the case locally 😅 
Maybe it's only due to my local data and can't happen in production 🤷 

I think that we can't be 100% sure document has children documents even in production so I prefer to handle the case where there's no children documents instead of adding the required directive. Do you agree ?

Comment (jpkha) [reply apps/portal-front/app/(public)/cybersecurity-solutions/[slug]/[docSlug]/slug-document.tsx]: Yes I agree 👍

PR #1229, opened 2025-11-12
[backend] fix boolean metadata which is not correctly convert

jpkha:
# Context

The value from document_metadata is stored as a string by default because of how it’s saved in the PostgreSQL table.
I’ve added a field for values that are known to be boolean. This setup is currently static, but we can make it dynamic in the future if we need more flexibility.

Verify that some connectors remain unverified (verify = false) and that some are undeployable (manager_supported = false).

## Related issues

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/portal-front/schema.graphql
+++ b/apps/portal-front/schema.graphql
@@ -560,12 +560,12 @@ type Connector implements Node & DocumentBase & IntegrationFeed {
   integration_type: IntegrationFeedType!
   product_version: String!
   container_image: String
-  verified: String!
+  verified: Boolean!
   source_code: String
   subscription_link: String
   integration_subtype: ConnectorType!
-  manager_supported: String!
-  playbook_supported: String!
+  manager_supported: Boolean!
+  playbook_supported: Boolean!
 }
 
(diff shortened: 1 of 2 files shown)


apps/portal-api/src/modules/services/integration-feeds/integration-feeds.graphql
```diff
@@ -75,12 +75,12 @@ type Connector implements Node & DocumentBase & IntegrationFeed {
...
   product_version: String!
   container_image: String
-  verified: String!
+  verified: Boolean!
   source_code: String
   subscription_link: String
   integration_subtype: ConnectorType!
-  manager_supported: String!
-  playbook_supported: String!
+  manager_supported: Boolean!
```
Kroustille: Since this property is now a boolean, you can change the condition using `manager_supported` in `shareable-resource-connector-slug.tsx` file 😁

Issue #1246, opened 2025-11-17
Ability to see an admin dashboard of trials in queued status


EllynBsc:
**Overall Description**
Ability to see a dashboard of trials in queued status

PR #1357, opened 2025-12-05
[backend/frontend] feature(freeTrials): new trials dashboards page #1246

carinelebas:
# Context
We need to have a new page to display a trials dashboard. 
- under free trials feature flags
- the cancel/reorder buttons currently do nothing

## How to test
Check the visual of the dashboard. Test search/filter/pagination
## Related issues

- Related to #1246

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

We don't return the same graphql object to platform and the front end (and don't want to, the token should never be returned to the front end). So I renamed `loadDeploymentRequests` to `loadPlatformDeploymentRequests` in the app layer and in graphql to clarify this difference.

### DIFF
--- a/apps/portal-front/src/components/trials/trials-tab.tsx
+++ b/apps/portal-front/src/components/trials/trials-tab.tsx
@@ -54,7 +54,7 @@ const trialsTabConfig: Record<
   }
 > = {
   [TrialsTabType.Cancelled]: {
-    statuses: [DeploymentRequestHubStatusEnum.CANCELED],
+    statuses: [DeploymentRequestHubStatusEnum.CANCELLED],
   },
   [TrialsTabType.Expired]: {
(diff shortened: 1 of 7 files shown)


apps/portal-front/src/components/trials/trials-tab.tsx
```diff
@@ -0,0 +1,371 @@
...
+              },
+            },
+          ]
+        : []),
+    ],
+    [t, type]
+  );
+
+  const queryData = useLazyLoadQuery<trialsListQuery>(TrialsListQuery, {
+    count: 100,
```
Kroustille: You can load initial query with `pageSize`. You will have to set `pageSize` to 100 in `trial-list-localstorage`

Kroustille (approved):

Issue #1178, opened 2025-10-27
🎯 Block competitors emails on trials manually


EllynBsc:
**Overall Description**
Block competitors emails on trials

PR #1377, opened 2025-12-10
[backend/frontend] feat(deployments): add domain blacklisting to prevent deployment creation (#1178)

jbanety:
# Context

This PR implements a **domain blacklisting feature** to prevent certain organizations from creating free trial deployment requests. This feature is controlled via a new configuration variable `domains_blacklist` (and env var `DOMAINS_BLACKLIST`) that allow to specify a comma-separated list of domains that should be blocked from requesting free trials.

## Backend changes
- Added `domains_blacklist` configuration variable (environment: `DOMAINS_BLACKLIST`)
- Added validation logic in `createDeploymentRequest` to check if organization domains are blacklisted
- Added new error code `CantRequestFreeTrial`
- Exposed `domains_blacklist` in GraphQL Settings query for frontend consumption

## Frontend changes
- Added `domains_blacklist` to Settings GraphQL fragment and query
- Implemented `isBlacklisted` logic in `useFreeTrial` hook to check if user's email domain is blacklisted
- Disabled "Start Trial" button when user is blacklisted

# How to test

1. Set the config variable in `apps/portal-api/config/local.json` : 
```json
{
"domains_blacklist": "test.com,example.com"
}
```
2. Create an organization with `test.com` as domain and an user `test@test.com`
3. Login and you should see a disabled Start a trial button

# Related issue

- #1178

### DIFF
--- a/apps/portal-api/src/thirdparty/auth0/auth0.util.test.ts
+++ b/apps/portal-api/src/thirdparty/auth0/auth0.util.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from 'vitest';
+import { removeEmptyGroups } from './auth0.util';
+import { Auth0UpdateUserRBACInstance } from './client';
+
+describe('removeEmptyGroups', () => {
+  it('should remove entries with empty groups arrays', () => {
+    const input: Auth0UpdateUserRBACInstance = {
+      user1: { groups: ['admin', 'user'] },
+      user2: { groups: [] },
+      user3: { groups: ['viewer'] },
+      user4: { groups: [] },
+    };
+
+    const expected: Auth0UpdateUserRBACInstance = {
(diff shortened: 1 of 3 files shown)


apps/portal-front/src/components/service/trial-instances/start-trial-button.tsx
```diff
@@ -51,7 +51,18 @@ export const StartTrialButton: React.FC<Props> = ({
   const t = useTranslations();
   const environment = useRelayEnvironment();
 
-  const { freeTrial } = useFreeTrial();
+  const { freeTrial, isBlacklisted } = useFreeTrial();
+
+  if (isBlacklisted) {
```
Kroustille: The condition is reversed: button must be displayed when user is not blacklisted

Issue #1163, opened 2025-10-22
🎯 Ability for users to reach out to sales when trial is started


EllynBsc:
**Overall Description**
Ability for users to reach out to sales when trial is started

PR #1378, opened 2025-12-10
[backend/frontend] feat(trials): contact us from detail page (#1163)

jbanety:
# Context
Ability to contact sales without an active trial.

## Related issues

- Related to #1163


apps/portal-front/src/components/service/trial-instances/contact-us-button.tsx
```diff
@@ -9,7 +9,7 @@ import { useState } from 'react';
 import { useMutation } from 'react-relay';
 
 interface ContactUsButtonProps {
-  variant: 'default' | 'gradient';
+  variant: 'default' | 'outline-primary' | 'gradient';
```
carinelebas: I think this part is already fixed in Valentin's PR: https://github.com/FiligranHQ/xtm-hub/pull/1371/files#diff-0fdb31f6110251be452183c6cc0d77925bb3f7c904545a7efcc9760262d387e8

Maybe you want to check which version to keep so as to avoid conflicts?

PR #1389, opened 2025-12-12
[backend] feature(freeTrials): handle freeTrials expiration #1385

carinelebas:
# Context
When end date is in the past, trials need to expire. We run a cronjob daily to expire these trials. 
For test purposes, it is currently set every 10 min. This will be updated before PR is merged.

## How to test
Create an active trial with an end date in the past:
- check trial gets expired  (hub_status expired + target_state removed)
- check mail is sent
- check telemetry event is sent

## Related issues

- Related to #1385

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

### DIFF
--- a/apps/portal-front/src/components/service/trial-instances/try-opencti-form.tsx
+++ b/apps/portal-front/src/components/service/trial-instances/try-opencti-form.tsx
@@ -35,9 +35,12 @@ export const tryOpenCTIFormSchema = z.object({
   job_title: z.enum(JOB_TITLES),
   activity_sector: z.enum(ACTIVITIES_SECTOR),
   use_case: z.enum(USE_CASES),
-  acceptTerms: z.boolean().refine((value) => value === true, {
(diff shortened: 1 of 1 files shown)


apps/portal-api/src/modules/services/deployments/deployments.domain.ts
```diff
@@ -148,6 +148,14 @@ export const DeploymentRequestDomain = {
       .first();
   },
 
+  loadExpiredTrials: async (): Promise<DeploymentRequest[]> => {
```
Kroustille:
```suggestion
  loadTrialsToExpire: async (): Promise<DeploymentRequest[]> => {
```

Because the trials are not already expired since the hub status is still Active. What do you think ?

Kroustille (approved):

PR #1398, opened 2025-12-16
ci: optimize CI/CD workflow and remove Nx overhead

jbanety:
# Context

Drastically optimize the CI/CD workflow by eliminating redundant rebuilds and unnecessary complexity.

# Major Changes

### 1. Multi-layer caching
- Yarn cache
- Docker Buildkit cache
- Pre-built images: Eliminates rebuilds in docker-compose

### 2. Test optimizations
- Front tests: Use `docker run` instead of `docker-compose`
- Test images: Pre-built and reused across jobs

### 3. Architecture simplification
- Nx removed: Replaced with pure Yarn workspaces

# Breaking Changes

- `npx nx run portal-api:dev` → `yarn workspace portal-api dev`
- All images are always built on PRs (no affected detection)


.github/scripts/get-affected-projects.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+set -e
```
hervyt: **nitpick**: You can be even more strict with `set -euo pipefail `
No ?

Issue #1506, opened 2026-01-14
Product testing - Seed some fake users for testing purposes


EllynBsc:
**Overall Description**

At the moment, I can only try feature envs with filigran users that are in Azure (due to this new mapping we're doing)

I'd like to be able to use some fake users for for testing purposes.

For instance, When I tested this issue:
https://github.com/orgs/FiligranHQ/projects/17/views/39?pane=issue&itemId=142675169&issue=FiligranHQ%7Cxtm-hub%7C1338

my fake users always got redirected to Azure login.

<img width="972" height="800" alt="Image" src="https://github.com/user-attachments/assets/9f194b17-db42-4ee0-8f86-ab551158801c" />
<img width="725" height="807" alt="Image" src="https://github.com/user-attachments/assets/567b9b93-3078-4c49-aa29-f0bd6023b486" />

PR #1535, opened 2026-01-20
[backend] feature(seed): create pending users on fenv and dev

carinelebas:
# Context
Create pending users on feature env and dev environment

## How to test
Check there are pending users on feature env and dev environments

## Related issues

- Related to #1506 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/portal-api/tests/test-utils.ts
```diff
@@ -75,12 +75,12 @@ export async function createUser({
   first_name = 'John',
   last_name = 'Doe',
   selected_organization_id = 'ba091095-418f-4b4f-b150-6c9295e232c4',
```
jpkha: Actually, I know I did this but it quite confusing by reading the function.
Should we call it Professional_Organization_Id ? something like that instead of selected but in fact we are not selected in case of pending

apps/portal-api/tests/test-utils.ts
```diff
@@ -106,32 +116,35 @@ export async function createUser({
...
+              user_id: userId,
+              organization_id: userId,
+            },
+          ]
+        : []),
       {
         user_id: userId,
         organization_id: userId,
       },
     ]);
```
jpkha: If it pending is does duplicate the relation user_id and organization_id right ?
Also it should be reversed I think ?

If pending -> do not add the Filigran id.
If not -> add Filigran id (selected_organization_id)

Kroustille (approved):

PR #1697, opened 2026-02-13
[frontend] - feat(OAEVtrials): Update banners and navbar #1631

hervyt:
# Context
> This PR add the new banner for both product (keeping the old banner when FF is not enabled) and add active trials information in nav bar 

## How to test
- Do not connect. 
- Observe the banner on public pages with and without the FF 
- Connect as user with ADMINISTRATE_ORGANIZATION or MANAGE_PLATFORM_REGISTRATION capabilities. 
- You should see the button "Start a Trial". 
-> If FF is enable : a dropdown to choose the product 
-> If FF is not enabled, the trial directly opens. 

- Connect as a user without capabilities 
- You should not see the button "Start a Trial" 
- Adapt the trials for them to be active : 
-> You should see the information about the trials in the nav bar. 

A lot of tests... See the vidéo

## Related issues

- Related to #1631 

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [X] Local manual tests

### DIFF
--- a/apps/portal-api/src/modules/services/registration/registration.domain.ts
+++ b/apps/portal-api/src/modules/services/registration/registration.domain.ts
@@ -173,28 +173,28 @@ export const registrationDomain = {
       })
       .where(function () {
         if (onlyActive) {
-          this.where(
-            'DeploymentRequest.hub_status',
-            '=',
-            DeploymentRequestHubStatus.Active
-          ).andWhere(function () {
-            this.whereNull('DeploymentRequest.id').orWhere(
+          this.whereNull('DeploymentRequest.id').orWhere(function () {
+            this.where(
               'DeploymentRequest.counts_in_orga_quota',
               '=',
               true
+            ).andWhere(
+              'DeploymentRequest.hub_status',
+              '=',
+              DeploymentRequestHubStatus.Active
             );
           });
         }
         if (onlyTrial) {
-          this.where(
-            'DeploymentRequest.type',
-            '=',
-            DeploymentRequestDeploymentType.Trial
-          ).andWhere(function () {
-            this.whereNull('DeploymentRequest.id').orWhere(
+          this.whereNull('DeploymentRequest.id').orWhere(function () {
+            this.where(
               'DeploymentRequest.counts_in_orga_quota',
               '=',
               true
+            ).andWhere(
+              'DeploymentRequest.type',
+              '=',
+              DeploymentRequestDeploymentType.Trial
             );
           });
         }


apps/portal-front/src/components/service/trial-instances/banner/start-trial-banner-button.tsx
```diff
@@ -0,0 +1,206 @@
...
+    commitCreateDeploymentRequestMutationMutation({
+      variables: {
+        input: {
+          ...valuesWithoutAcceptTerms,
+          platform_identifier: platformIdentifier,
+          type: DeploymentRequestDeploymentTypeEnum.TRIAL,
+        },
+      },
+      updater: (store) => {
+        store.invalidateStore();
```
carinelebas: do we really need to invalidateStore if we refetch ?

apps/portal-front/src/components/service/trial-instances/useFreeTrials.ts
```diff
@@ -3,27 +3,62 @@ import {
...
-      .some((domain) => me?.email?.includes(domain.trim()));
+export type DeploymentRequest = {
+  activity_sector: string | null | undefined;
+  counts_in_orga_quota: boolean;
+  hub_status: DeploymentRequestHubStatus;
+  id: string;
+  job_title: string | null | undefined;
+  type: DeploymentRequestDeploymentType;
+};
+export type FreeTrial = {
```
carinelebas: question: why do you define this type instead of using the Fragment?

apps/portal-api/src/modules/services/registration/registration.app.ts
```diff
@@ -115,6 +115,8 @@ export const registrationApp = {
     const platforms = await registrationDomain.loadRegisteredPlatforms({
       platformIdentifier: input?.identifier,
       onlyActive: input?.onlyActive ?? false,
+      onlyTrial: input?.onlyTrial ?? false,
+      onlyCountsInOrgaQuota: input?.onlyCountsInOrgaQuota ?? false,
```
carinelebas: I don't think we need to add this parameter, I don't see any use case we we need to return to the front end a platform that doesn't count in orga quota. Maybe if specifically requested by serviceInstanceId (and even then? ), but otherwise, they are not supposed to exist. it's a kind of soft delete. So I think we could always add in the query counts_in_orga_quota=false.

Kroustille (approved):

apps/portal-front/messages/en.json
```diff
@@ -871,12 +871,19 @@
...
+      "ExploreProducts": "Explore OpenCTI or OpenAEV platform with 30 days",
       "Active": "Your OpenCTI free trial is active",
       "Provisioning": "Your OpenCTI free trial is currently being provisioned.",
       "Requested": "Your OpenCTI free trial has been requested.",
       "Expired": "Your OpenCTI free trial has expired.",
       "Cancelled": "Your OpenCTI free trial has been cancelled.",
       "DaysRemaining": "{days, plural, =1 {# Day remaining} other {# Days remaining}}",
+      "Header": {
+        "ActiveTrial": "Active trial",
+        "DaysRemaining": "remaning day"
```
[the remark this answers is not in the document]
hervyt: The s is displayed in the component depending the number of remaning days ;)

apps/portal-api/src/modules/services/registration/registration.app.ts
hervyt: I did remove it, but in the query, counts_in_orga_quota=true, no ? 
Before my PR in the front, we filtered with this param to true :thinking:

carinelebas (approved):

Issue #1978, opened 2026-03-30
[bug] cancellation reason display in trial dashboard


carinelebas:
**Describe the bug**
Cancellation reason in trial dashboard is not well displayed with 'Other': it's displayed `Service.Trials.CancellationReason.Other : <free text>` 
It should be `Other: <free text>`

**To Reproduce**

**Actual behavior**
it's displayed `Service.Trials.CancellationReason.Other : <free text>` 

**Expected behavior**
It should be `Other: <free text>`

PR #2002, opened 2026-03-31
[frontend] fix(trials): display cancellation reason correctly for Other with free text (#1978)

jbanety:
# Context

The trial dashboard was incorrectly displaying cancellation reasons when a user selected **"Other"** and typed free text. Instead of showing `Other: <free text>`, the UI was showing the raw i18n key path `Service.Trials.CancellationReason.Other : <free text>` because `next-intl` couldn't find a matching translation key for the compound string.

The root cause: the display code blindly passed the raw DB value to `t()` as a translation key suffix. When the stored value contains free text (e.g. `Other: my custom reason`), no matching key exists and `next-intl` falls back to the full key path.

The fix detects all "Other" variants and formats them correctly using the already-localized `CancellationReasonOther` label:

| DB value | Before | After |
|---|---|---|
| `Other: some text` | `Service.Trials.CancellationReason.Other: some text` | `Other: some text` |
| `Other` | `Service.Trials.CancellationReason.Other` | `Other` |
| `other` (migrated) | `Service.Trials.CancellationReason.other` | `Other` |

## How to test

Capture d’écran 2026-03-31 à 09 52 06

1. As an admin, go to the **Trials Dashboard**
2. Cancel a trial and select **"Other"** as the cancellation reason, then type some free text (e.g. `my custom reason`) and confirm
3. Back on the dashboard, check the **Cancellation Reason** column for that trial
4. It should display `Other: my custom reason` instead of `Service.Trials.CancellationReason.Other: my custom reason`

Also verify existing trials with pre-migrated `other` (lowercase) or plain `Other` values display `Other` correctly.

## Related issues

- Related to #1978

# Checklist

## Quality

## Testing

- [x] (Bug fixes only) I added tests that reproduce and verify the fix

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests


hervyt (approved):
💯

Issue #2034, opened 2026-04-02
[tech] Use ID directive for xxx_id fields in graphql


carinelebas:
In many places in the code, we need to encode/decode relay id: when the field is not named id, it's not automatically done. 
We could use a `@id` directive to make this encoding/decoding, and cast to the right type.

carinelebas:
discarded in favor of https://github.com/FiligranHQ/xtm-hub/issues/2039

PR #2035, opened 2026-04-02
[backend] feat: add directive to encode/decode relay id (#2034)

carinelebas:
# Context
in order to avoid manually encoding/decoding relay ids, we can use a `@id` directive that will automatically manage this, and cast to the right type. 
This can only apply to Id that are not directly the node `ID`. 
Implement a first use case on deployment Request. 

- test there is no regression on deployment Requests.

## Related issues

- Related to #2034

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/portal-api/src/modules/services/deployments/deployments.graphql
+++ b/apps/portal-api/src/modules/services/deployments/deployments.graphql
@@ -1,3 +1,5 @@
+scalar DeploymentRequestId
+
 enum DeploymentRequestPlatformRegion {
   eu_west
   us_east
@@ -119,8 +121,8 @@ type DeploymentRequest implements Node {
   hub_status: DeploymentRequestHubStatus!
   ordering: Int!
   requester_email: String
-  organization_requester_id: ID!
-  service_instance_id: ID!
+  organization_requester_id: OrganizationId! @id(type: "Organization")
+  service_instance_id: ServiceInstanceId! @id(type: "ServiceInstance")
   organization_name: String
   request_date: Date!
   counts_in_orga_quota: Boolean!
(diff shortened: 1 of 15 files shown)


apps/portal-api/src/modules/services/deployments/deployments.graphql
```diff
@@ -119,8 +121,8 @@ type DeploymentRequest implements Node {
   hub_status: DeploymentRequestHubStatus!
   ordering: Int!
   requester_email: String
-  organization_requester_id: ID!
-  service_instance_id: ID!
+  organization_requester_id: OrganizationId! @id(type: "Organization")
```
jpkha: That's a good idea to use scalar for defining specific id.

But I think the directive might be not the good solution.

I would say create a custom scalar and declare in the Apollo might be the best option :
https://www.apollographql.com/docs/apollo-server/schema/custom-scalars#providing-custom-scalars-to-apollo-server

Actually you can define the scalar transformation before the resolver and after.

It might resolve multiple issue we face like transforming manually before sending a response but also when receiving it !

carinelebas: Actually, the directive is only use for encoding/decoding the id. The type cast with the scalar is done in [codeGen](https://github.com/FiligranHQ/xtm-hub/blob/4bc4e966024a9fa7bd10aa32deb89821ca6d82c1/apps/portal-api/codegen.yml#L16)

jpkha (changes requested):
This PR can be closed. Another solution was adopted instead

carinelebas:
replaced by https://github.com/FiligranHQ/xtm-hub/pull/2040

PR #2050, opened 2026-04-03
Chart cnpg bareman plugin

pierremahot:
# chart updates
update cnpg plugin usage


chart/templates/postgres.yml
```diff
@@ -15,7 +15,15 @@ spec:
...
-      database: xtmhub
-      owner: {{ .Values.postgres.username }}
+    {{- .Values.postgres.spec.bootstrap | toYaml | nindent 4 }}
+  {{- end }}
+  {{- if .Values.postgres.spec.externalClusters }}
+  externalClusters:
+    {{- .Values.postgres.spec.externalClusters | toYaml | nindent 4 }}
+  {{- end }}
+  {{- if .Values.postgres.spec.plugins }}
+  plugins:
```
victorhery-filigran, line 27: When backup are enabled, don't we need the backup's plugin enabled too?

chart/templates/postgres.yml
```diff
@@ -15,7 +15,15 @@ spec:
   imagePullPolicy: "IfNotPresent"
   primaryUpdateMethod: switchover
   primaryUpdateStrategy: unsupervised
+  {{- if .Values.postgres.spec.bootstrap }}
   bootstrap:
-    initdb:
-      database: xtmhub
-      owner: {{ .Values.postgres.username }}
+    {{- .Values.postgres.spec.bootstrap | toYaml | nindent 4 }}
```
victorhery-filigran, line 20: Just to be sure, moving this to values.yaml is not related to backup itself, as default option you set are the same than actual?
Asking because of the possibility to bootstrap from backup instead of master with the new backup plugin :)

victorhery-filigran (commented):
Some questions about the backup concept to be sure I don't miss anything :-)

Thanks a lot for the work!

chart/templates/postgres.yml
victorhery-filigran: As discussed, we want to be able to surcharge this parameter during backup, so having it in values is perfect 👌

victorhery-filigran (approved):

Issue #2121, opened 2026-04-14
[backend] (tests): improve tests with helpers


hervyt:
Today, a lot of API tests directly call the DB in the tests while helpers exist. 

We should rather call the helpers. We should add an ESLint rule to avoid directly calling DB

PR #2130, opened 2026-04-14
[backend] chore(TestsApi): Add Liinters rules (#2121)

hervyt:
# Context
> Improve and refacto API Tests

## How to test
- All API tests should pass

## Related issues

- Related to #2121

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/portal-api/package.json
```diff
@@ -63,6 +63,8 @@
     "body-parser": "2.2.2",
     "config": "4.4.1",
     "cors": "2.8.6",
+    "eslint-plugin-jest": "29.15.2",
```
jpkha, line 66: it's weird, we should not used jest but vitest instead ?

hervyt: Yep, but we can use the jest plugin for ESLint because it's more complete in the rules (mooooore rules :D). We can combine both :)

apps/portal-api/src/modules/deployment/competitor/competitor.app.test.ts
```diff
@@ -20,9 +20,9 @@ const DEFAULT_COMPETITOR = {
 const createCompetitor = (overrides: Partial<CompetitorInitializer> = {}) =>
   CompetitorDomain.insertCompetitor({ ...DEFAULT_COMPETITOR, ...overrides });
 
-describe('CompetitorApp', () => {
+describe('competitorApp', () => {
```
Kroustille, line 23: I do not completely agree with the rule telling that describes should start with a lowercase, because here we losing the fact that we're testing `CompetitorApp`. 

I agree that `it` should start with a lowercase but maybe not the describes.

What do you think ? 🤔

apps/portal-api/tests/helper/test.document.helper.ts
```diff
@@ -0,0 +1,149 @@
...
+  encoding: '7bit',
+  createReadStream: vi.fn(),
+};
+
+const mockUpload = {
+  file: mockFileUpload,
+  promise: Promise.resolve(mockFileUpload),
+};
+
+export const TestDocumentHelper = {
```
Kroustille, line 38: We can split test helper in different helpers, but still use only the TestHelper in order to have only one way to create anything we want in tests.

To do so, we could destructure TestDocumentHelper into TestHelper 😉

apps/portal-api/src/modules/deployment/competitor/competitor.app.test.ts
hervyt: I added this rule because in our describe, it, etc. we just display  the function name. Or, a philosophy in jest/vitest is that: "tests as readable behavior descriptions, not titles". This is why this rule exists in linter (and I add it) and it helps us to provide clear and human readable description (and today, it not detailed, nor human readable!) .

Kroustille: Okay so if I understand well in this philosophy, we should call this describe `competitor app` right ? 

I was thinking that `CompetitorApp` is still better because we directly know what we're testing/what is failing 😛

hervyt: Yep indeed ! We should describe what it competitior app doing, so it's easier to read. However, did not do it here, because the PR is large enought....

Kroustille: For sure 😉 
Did you plan to talk about this new ES lint rules in tech pain to be sure everyone is up to date with test naming ? 🙂

apps/portal-api/eslint.config.mjs
```diff
@@ -35,9 +53,36 @@ export default defineConfig([
               message:
                 'Use a dedicated test context helper instead of bypass/admin shared contexts.',
             },
+            {
+              group: ['**/tests/helper/test.document.helper'],
```
Kroustille: Do you think it's possible to use a rule with a group like `'**/tests/helper/test.*.helper'` to be sure we don't have to think about adding a new rule for each new test helper ? 🤔

Kroustille (approved):

PR #2148, opened 2026-04-15
[backend] feat: create and provision news feed item on dashboard publish (#2134)

Kroustille:
# Context
>

- create an unpublished custom dashboard
- check that no news feed item is created
- create a published custom dashboard
- check that news feed item and provisioned news feed item are created
- unpublish and then publish this dashboard
- check that news feed item and provisioned news feed item are created

## Related issues

- Related to #2134

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/portal-api/src/modules/document/document.app.ts
```diff
@@ -326,8 +355,27 @@ export const DocumentApp = {
...
+      documentBeforeUpdate.active === false &&
+      updatedDocument.active === true &&
+      NewsFeedApp.isNewsFeedConfigured(serviceDefinition.identifier)
+    ) {
+      NewsFeedApp.createResourceNewsFeedItem({
+        document: updatedDocument,
+        serviceInstanceId,
+        serviceDefinitionIdentifier: serviceDefinition.identifier,
+      }).catch((error) =>
+        logApp.error('Unable to create news feed item on dashboard update', {
```
jpkha: Are we only checking dashboard? Isn't any document ? Maybe if we can target directly all the document we have for newsfeed instead of dashboard, it would be better. More business value and less time and complexity to manage.

jpkha (approved):

apps/portal-api/src/modules/document/document.app.ts
Kroustille: For the moment we're only adding news feed for dashboards but we will surely add other types of feed in the future.
We're only implementing feed for OpenCTI at this moment, and I think that integrations needs to be displayed in another way because of their integration type so we will need to brainstorm it before implementation.

I'm updating the log to be generic instead of focusing on dashboards only 😉

PR #2175, opened 2026-04-17
[backend/frontend] refactor(HomePage): Simplify HomepageList (#2157)

hervyt:
# Context
> We aim at simplify homepage and subscription behaviour. Now, the user is subscribed automatically to the service when clicking on the tile. Every user of organization can do that.
This PR will be merged on the feature branch that will be tested at the end.  

## How to test
- Connect on random user on the Hub 
- Click on a service tile
- You should see the subscription in the table, and then the user_service
- Connect as another user from the same organization 
- Click on the same service tile 
- The subscription should not be added (already exists) but the user_service should be added. 

## Related issues

- Related to 2157 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/portal-front/src/components/service/service-list.tsx
```diff
@@ -1,150 +1,88 @@
...
-    toast({
-      variant: 'destructive',
-      title: t('Utils.Error'),
-      description: <>{t(`Error.Server.${error.message}`)}</>,
-    });
-  };
+  const getRegisteredPlatforms = (
+    queryRef: PreloadedQuery<registerRegisteredPlatformsQuery>
+  ) => {
+    const queryData = usePreloadedQuery<registerRegisteredPlatformsQuery>(
```
Kroustille: I think you should move it into a `useCallback` or `useMemo`, since it's forbidden to conditionnaly call React hooks 🤔 

Maybe we can also extract the whole method into a new custom hook

apps/portal-front/src/components/service/document/use-document-context.ts
```diff
@@ -259,13 +261,22 @@ export function useDocumentContext({
     return translationKeyMapping[type]();
   }, [type, integrationType]);
 
+  const currentUserSubscriptionId = me?.id
+    ? serviceInstance.subscriptions?.find((subscription) =>
```
Kroustille, line 265: I'm wondering : if this find is required, do we fetch subscriptions for every users in the query ? Is it filtered on the backend ?

apps/portal-api/src/modules/service/instance/service-instance.domain.ts
```diff
@@ -259,19 +222,34 @@ export const loadServiceInstances = async (opts) => {
...
+      'Subscription.organization_id'
+    );
+
+  if (!isUserAdminPlatform(user)) {
+    queryBuilder.where(
+      'Subscription.organization_id',
+      '=',
+      user.selected_organization_id
+    );
+  }
```
Kroustille, line 241: You can use `.tap(restrictSubscriptionToUserOrganization)` or a similar call

apps/portal-front/src/components/service/document/use-document-context.ts
hervyt: Only BYPASS has vveyrthing, otherwise it's still filtered in the backend. I added this find for safety :wink:

apps/portal-api/src/modules/service/instance/service-instance.domain.ts
hervyt: Can cannont, since the joins are not the same ! :scream:  (I thought it was OK too, but it fails (tests and manual tests) (we go from Subscription and not user service like in the function you proposed)

Kroustille: Ho yes I see, and since the `leftJoin` is already in `loadServiceInstanceSubscriptions`, it won't be useful to add a restrictions there 👍

apps/portal-front/src/components/service/service-list.tsx
```diff
@@ -1,150 +1,85 @@
...
-    queryRefUserServiceOwned,
-    queryRefServiceList,
-    queryRefRegisteredPlatforms,
-    onUpdate,
-    handleSuccess,
-    handleError
+  const registeredPlatforms = useMemo(
+    () => registeredPlatformsData.registeredPlatforms ?? [],
+    [registeredPlatformsData.registeredPlatforms]
   );
```
Kroustille: I think we could replace this useMemo by adding a `useMemo` around 

```
const sortedServices = [
    ...freeTrialsSkeletonDataCards,
    ...registeredPlatforms.map((platform) =>
      registeredPlatformToServiceInstanceCardData(platform, t)
    ),
    ...serviceData.map(publicServiceInstanceToInstanceCardData),
  ].sort((a, b) => a!.ordering - b!.ordering);
```

because when declaring sortedServices, we're destructuring multiple arrays and it would be better for performance to have everything in a `useMemo`.

Please tell me if it's not clear 😅

Kroustille (approved):

PR #2162, opened 2026-04-16
[backend/frontend] feat: Register/Unregister with tenant (#2126)

carinelebas:
# Context
We need to support different tenant on the same platform. 
This PR handled registration/unregistration with a tenant. 

Related OpenAEV PR is:  https://github.com/OpenAEV-Platform/openaev/pull/5503

## How to test

- Register/unregister different openAEV tenant: 
  - Check registration/unregistration works: platform is well displayed/removed in frontend. Link to the platform is well the one provided
  - Check telemetry event contains the tenant id
- Check error cases
  - try to register an openaev platform with version < 2.4.0 and no tenant provided: it should not work 

## Related issues

- Related to #2126

# Checklist

## Quality

- [x] I consider this work complete and ready for review
- [x] I tested all features and edge cases
- [ ] I refactored code where necessary to improve quality
- [ ] I made sure my code meets development guidelines
- [x] I performed a self-review of my code
- [ ] I added the `skip-feature-env` label if a feature environment is not needed for this PR

## Testing

- [x] I wrote test cases for the relevant use cases (unit, integration or e2e tests)
- [ ] (Bug fixes only) I added tests that reproduce and verify the fix

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information
> Any additional context, dependencies, or concerns reviewers should know about.

PR #2183, opened 2026-04-20
[frontend/backend] fix(import-from-hub): support tenant_id and accept oaev_instance_id alias (#2145)

Antoine-Quintin:
# Context
>

Part of the OpenAEV multi-tenancy epic. This PR adapts the "Import from Hub" flow to support multi-tenant OpenAEV instances, building on the foundations laid by #2162 (register/unregister with tenant).

**Backend changes:**
- `platformAssociatedOrganization` GraphQL query now accepts an optional `tenantId` argument
- `loadPlatformAssociatedOrganization` propagates `tenantId` down to `loadConfigurationByPlatform` (already multi-tenant aware since #2162)

**Frontend changes:**
- `/redirect/:identifier` resource flow now reads `tenant_id` from query params and propagates it through the resolution chain (`resource.ts` → `loadPlatformOrganizationId` → Relay query)
- Added backwards-compatible alias: `oaev_instance_id` is accepted as a synonym of `platform_id` to support OpenAEV instances that haven't migrated to the new query param name yet (same pattern as the existing `opencti_platform_id` → `platform_id` alias)
- **Redirect fallback fix** : when `platform_id`/`tenant_id` don't match any registered organization, the redirect used to silently fall back to the user's personal space. It now falls back to the user's currently selected organization, so users stay in their working context when "Import from Hub" is triggered from an unregistered platform.

## Related issues

- Related to #2145 
- Companion OpenAEV PR: OpenAEV-Platform/openaev#5529

# Checklist

## Quality

## Testing

- [x] (Bug fixes only) I added tests that reproduce and verify the fix

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

**Commits** (kept separate for easier review):
1. `fix(import-from-hub): support tenant_id and accept oaev_instance_id alias` — the main change covering both back and front
2. `test(registration): cover tenantId forwarding in platformAssociatedOrganization resolver` — unit test for the resolver
3. `fix(redirect): fallback to user selected organization instead of personal space when platform has no associated organization` — additional UX fix (see "Redirect fallback fix" above)

### DIFF
--- a/apps/portal-api/src/modules/registration/registration.app.test.ts
+++ b/apps/portal-api/src/modules/registration/registration.app.test.ts
@@ -9,7 +9,9 @@ import {
   it,
   vi,
 } from 'vitest';
-import { TestHelper } from '../../../tests/helper/test.helper';
+import {
+  TestHelper,
+} from '../../../tests/helper/test.helper';
 import {
   // eslint-disable-next-line no-restricted-imports
   contextBypassUser,
@@ -91,6 +93,7 @@ describe('registration app', () => {
       await TestHelper.serviceInstance.create({
         id: serviceInstanceId,
         name: 'test',
+        service_definition_id: SERVICES.DEFINITIONS.OPENCTI_REGISTRATION.ID,
       });
       await TestHelper.serviceConfiguration.create({
         service_instance_id: serviceInstanceId,
@@ -111,6 +114,7 @@ describe('registration app', () => {
       await TestHelper.serviceInstance.create({
         id: serviceInstanceId,
         name: 'test',
+        service_definition_id: SERVICES.DEFINITIONS.OPENCTI_REGISTRATION.ID,
       });
       await TestHelper.serviceConfiguration.create({
         service_instance_id: serviceInstanceId,
@@ -129,6 +133,27 @@ describe('registration app', () => {
       await expect(call).rejects.toThrow(ErrorCode.UserIsNotInOrganization);
     });
 
+    it('should throw TenantIdMandatory when platform requires tenantId and none is provided', async () => {
+      const platformId = uuidv4();
+      const serviceInstanceId = uuidv4() as ServiceInstanceId;
+
+      // Use OpenAEV service definition (OpenAEV requires tenantId from v2.4.0)
+      await TestHelper.serviceInstance.create({
+        id: serviceInstanceId,
+        name: 'test',
+        service_definition_id: SERVICES.DEFINITIONS.OPENAEV_REGISTRATION.ID,
+      });
+      await TestHelper.serviceConfiguration.create({
+        service_instance_id: serviceInstanceId,
+        config: { platform_id: platformId, platform_version: '2.5.0' },
+        status: ServiceConfigurationStatus.Active,
+      });
+
+      const call =
+        registrationApp.loadPlatformAssociatedOrganization(platformId);
+      await expect(call).rejects.toThrow(BadRequestErrorCode.TenantIdMandatory);
+    });
+
     it('should return an organization when platform is found and user is allowed', async () => {
       const platformId = uuidv4();
       const serviceInstanceId = uuidv4() as ServiceInstanceId;
@@ -136,6 +161,7 @@ describe('registration app', () => {
       await TestHelper.serviceInstance.create({
         id: serviceInstanceId,
         name: 'test',
+        service_definition_id: SERVICES.DEFINITIONS.OPENCTI_REGISTRATION.ID,
       });
       await TestHelper.serviceConfiguration.create({
         service_instance_id: serviceInstanceId,
--- a/apps/portal-api/src/modules/registration/registration.app.ts
+++ b/apps/portal-api/src/modules/registration/registration.app.ts
@@ -66,11 +66,14 @@ import { ServiceConfigurationDomain } from './service-configuration/service-conf
 
 export const registrationApp = {
   loadPlatformAssociatedOrganization: async (
-    platformId: string
+    platformId: string,
+    tenantId?: string | null
   ): Promise<Organization | null> => {
     const { user } = requestContext.require();
     const serviceConfiguration =
-      await ServiceConfigurationDomain.loadConfigurationByPlatform(platformId);
+      await ServiceConfigurationDomain.loadConfigurationByPlatform(platformId, {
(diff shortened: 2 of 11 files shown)


apps/portal-api/src/modules/registration/registration.app.ts
```diff
@@ -81,11 +81,14 @@ const isTenantIdRequired = (
...
 export const registrationApp = {
   loadPlatformAssociatedOrganization: async (
-    platformId: string
+    platformId: string,
+    tenantId?: string | null
   ): Promise<Organization | null> => {
     const { user } = requestContext.require();
     const serviceConfiguration =
-      await ServiceConfigurationDomain.loadConfigurationByPlatform(platformId);
+      await ServiceConfigurationDomain.loadConfigurationByPlatform(platformId, {
```
carinelebas, line 74: What happens if we have several tenants registered on the same platform, but we don't provide an tenantId?
I think we should check if tenant is required after we retrieve the configuration, like [here](https://github.com/FiligranHQ/xtm-hub/blob/3a1955c7a3b657ba6a6b421db04a6b6351066719/apps/portal-api/src/modules/registration/registration.app.ts#L219). 
Wdyt?

apps/portal-front/app/redirect/[identifier]/resource.ts
```diff
@@ -90,7 +90,12 @@ export const redirectToResource = async (
     )!.id;
 
     // 3. Switch to the found organization
-    await switchOrganization(organizationId ?? personalSpaceGlobalId);
+    // If no organization is associated with the platform/tenant, fall back to the user's
+    // currently selected organization instead of their personal space, so users stay in
+    // their working context when Import from Hub is triggered from an unregistered platform.
+    await switchOrganization(
+      organizationId ?? user.selected_organization_id ?? personalSpaceGlobalId
```
Comment (carinelebas) [inline apps/portal-front/app/redirect/[identifier]/resource.ts]: `user.selected_organization_id` is of type string, I don't think it can be undefined, so `?? personalSpaceGlobalId` is unreachable, no?

Comment (Antoine-Quintin) [reply apps/portal-front/app/redirect/[identifier]/resource.ts]: You're right about the type - TypeScript says it's always a string. I kept the 
fallback as a defensive safety net in case of unexpected runtime issues (backend 
changes, stale GraphQL cache, etc.). Happy to remove it if you prefer the code 
to strictly follow the type - let me know!

Comment (carinelebas) [reply apps/portal-front/app/redirect/[identifier]/resource.ts]: I'd rather simplify the code and keep it aligned with scenarii that can really happen. 
if something unexpected happens, it should rather be treated in an exception flow (so with a try catch, if we really want to catch it), rather than with a silent fallback, with which we wouldn't even know something is not working well.

apps/portal-api/src/modules/registration/registration.app.ts
```diff
@@ -78,6 +78,25 @@ export const registrationApp = {
...
+    }
+    const config = serviceConfiguration.config as PlatformConfiguration;
+    const platformIdentifier =
+      platformIdentifierMappedByServiceDefinitionIdentifier[
+        serviceDefinition.identifier
+      ];
+    if (
+      platformIdentifier &&
+      isTenantIdRequired(platformIdentifier, config.platform_version) &&
+      !tenantId
```
carinelebas: - It would be better to make the check of  !tenantId before loading the serviceDefinition. If the tenantId is provided, this will avoid one DB call. 
- if we don't have a platformIdentifier here, it means we have a big problem with the relations of our entities. We should rather throw an exception than skip this test.

Note: it is better to evaluate the cheapest conditions first, because the right side of the expression is not evaluated if the left part is already false. In this case, `isTenantRequired` is (just a bit) more expensive that `!tenantId`, so it would have been better to put it first (And also imo clearer :) )

apps/portal-api/src/modules/registration/registration.app.test.ts
```diff
@@ -129,6 +132,34 @@ describe('registration app', () => {
...
+        identifier: ServiceDefinitionIdentifier.OpenaevRegistration,
+      });
+      await TestHelper.serviceInstance.create({
+        id: serviceInstanceId,
+        name: 'test',
+        service_definition_id: serviceDefinition.id,
+      });
+      await TestHelper.serviceConfiguration.create({
+        service_instance_id: serviceInstanceId,
+        config: JSON.stringify({
```
carinelebas: you don't need JSON.stringify (see other similar test cases)

apps/portal-api/src/modules/registration/registration.app.test.ts
```diff
@@ -108,9 +114,13 @@ describe('registration app', () => {
       const platformId = uuidv4();
       const serviceInstanceId = uuidv4() as ServiceInstanceId;
       const subscriptionId = uuidv4() as SubscriptionId;
+      const serviceDefinition = await TestHelper.serviceDefinition.create({
+        identifier: ServiceDefinitionIdentifier.OpenctiRegistration,
+      });
       await TestHelper.serviceInstance.create({
         id: serviceInstanceId,
         name: 'test',
+        service_definition_id: serviceDefinition.id,
```
carinelebas: you don't need to create a new service definition, you can directly use `SERVICES.DEFINITIONS.OPENCTI_REGISTRATION.ID`

apps/portal-api/src/modules/registration/registration.app.test.ts
```diff
@@ -129,13 +139,41 @@ describe('registration app', () => {
...
+      const serviceInstanceId = uuidv4() as ServiceInstanceId;
+
+      // Create an OpenAEV service definition (OpenAEV requires tenantId from v2.4.0)
+      const serviceDefinition = await TestHelper.serviceDefinition.create({
+        identifier: ServiceDefinitionIdentifier.OpenaevRegistration,
+      });
+      await TestHelper.serviceInstance.create({
+        id: serviceInstanceId,
+        name: 'test',
+        service_definition_id: serviceDefinition.id,
```
carinelebas: Here it would be better to add the `OPENAEV_REGISTRATION` service definition id in the `tests.const.ts` file, like it's done for `OPENCTI_REGISTRATION` instead of creating a new service definition

carinelebas (approved):
👍

PR #2203, opened 2026-04-22
[backend] feat(refreshConnectivity): new endpoints for multi-tenancy (#2143)

carinelebas:
# Context
- create 2 new endpoints to check connectivity for multi tenants.
  - to refresh a single tenant
  - to refresh all tenants from the platform
- the refresh connectivity will also be used to migrate from platform with no-tenant, to a platform with tenants:  
  - updates the url
  - updates the version
 
## How to test
- register the same platform for multiple tenants:
  - Unregister one from the hub.
  - Connectivity Check single tenant must check the right tenant configuration
  - Connectivity Check multiple tenant must:
    - validate all tenant
    - send back the right status for each tenant
    - must set as inactive any tenant that wasn't sent in input. 
- register an openaev platform without tenant with version < 2.4.0
  - call connectivityCheck with tenant and version >= 2.4.0: check must be ok, version and url must be updated
- register an openaev plaform with tenant and version >= 2.4.0
  - connectivityCheck without tenant (old enpoint) must fail  

## Related issues

- Related to #2143
- Linked to https://github.com/OpenAEV-Platform/openaev/pull/5550

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [x] E2E tests
- [x] Local manual tests

# Additional information


apps/portal-api/src/modules/registration/service-configuration/service-configuration.domain.test.ts
```diff
@@ -110,13 +110,48 @@ describe('serviceConfigurationDomain', () => {
...
-            platformId: platformId(),
+            platform_id: platformId(),
             token: token(),
           });
 
         expect(configuration).toBeUndefined();
       }
     );
+
+    describe('withoutTenantId option', () => {
```
Kroustille, line 121: I don't find test cases where this option is false or not provided and configuration has tenant id 🤔

apps/portal-api/src/modules/registration/registration.app.test.ts
```diff
@@ -1167,6 +1181,420 @@ describe('registration app', () => {
...
+        const updatedConfig =
+          await ServiceConfigurationDomain.loadConfigurationByPlatform(
+            platformId,
+            { tenantId }
+          );
+        expect(updatedConfig?.config).toMatchObject({ tenant_id: tenantId });
+      });
+    });
+
+    describe('refreshPlatformRegistrationConnectivityStatusMultipleTenants', () => {
```
Kroustille: Is there a difference between single and multiple tenants mutations ? 
Isn't it possible to only have the multiple one called with an array containing only one tenant ?

apps/portal-api/src/modules/registration/registration.app.ts
```diff
@@ -149,59 +153,81 @@ export const registrationApp = {
...
+            `Failed to deactivate stale configuration for service instance ${config.service_instance_id}`,
+            { error }
+          );
+        })
+      )
+    );
+    return { statuses };
+  },
+
+  refreshPlatformRegistrationConnectivityStatus: async (
```
Kroustille: We begin to have quite a lot of code dedicated to connectivity refresh. Maybe we could move it in a new registration app ? 😄

apps/portal-api/src/modules/registration/registration.app.ts
```diff
@@ -615,3 +641,110 @@ const assertValidDeploymentRequest = (
     throw new Error(ForbiddenErrorCode.NotAllowedByDeploymentStatus);
   }
 };
+
+const refreshConnectivityStatus = async ({
```
Kroustille: This function begin to be hard to read. We could refactor it a bit, put computation in other functions and simplify condition readability by moving them to new constants

apps/portal-api/src/modules/registration/registration.app.ts
carinelebas: done in [535a191](https://github.com/FiligranHQ/xtm-hub/pull/2203/commits/535a191eec32ebd0329441de2fff7e201c3261e7)

apps/portal-api/src/modules/registration/registration.app.ts
carinelebas: done in [ab71723](https://github.com/FiligranHQ/xtm-hub/pull/2203/commits/ab71723e6eef25fc88fd0019ff5444c7b2dc1760)

apps/portal-api/src/modules/registration/service-configuration/service-configuration.domain.test.ts
carinelebas: good point! done in [ab71723](https://github.com/FiligranHQ/xtm-hub/pull/2203/commits/ab71723e6eef25fc88fd0019ff5444c7b2dc1760)

apps/portal-api/src/modules/registration/registration.app.test.ts
carinelebas: The all tenants endpoint updates all tenants, but also deactivate all tenants of the platform that are not provided in input. So it clearer and safer to have 2 different endpoints.

Kroustille (approved):
I find it clearer this way, thanks a lot 😁

Issue #2115, opened 2026-04-13
Add Japanese language in XTM Hub


jpkha:
**Overall Description**
A clear and concise description of what the feature is about. The scope should be clearly defined.

We need to align with Filigran website to support the same language.

**Use case**
Describe the use case for which you need a solution

Add Japanese translation, we can start with a automatic translation.

We need to setup the public path for Japanese and parse all the public path to add next-intl for translation.

Something like : 
- http://hub.filigran.io/ja/ 
- http://hub.filigran.io/fr/ 
- http://hub.filigran.io/de/ 

**Additional context**

PR #2238, opened 2026-04-27
[frontend] feat(i18n): add Japanese locale and prefixed public URLs (#2115)

Julienger0:
# Context
>
image

- Add Japanese (`ja`) as a supported locale alongside English and French 
- Surface Japanese in the profile language dropdown 
- yarn auto-translation now translates en to fr and en to ja in one pass 
- Public routes move under app/[locale]/(public)
- SEO localization: per-locale (en,ja)
- only japanese and english are available on public pages
- Metadata strings externalized into the Metadata namespace in i18n message files 
- add slug to service instances who didn't have on
- add translations

## Related issues

- Related to #2115 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [x] E2E tests
- [ ] Local manual tests

# Additional information


apps/portal-front/app/[locale]/(public)/cybersecurity-solutions/[slug]/[docSlug]/page.tsx
```diff
@@ -89,19 +99,25 @@ export async function generateMetadata({
     document.id
   );
 
+  const pathname = `/${PUBLIC_CYBERSECURITY_SOLUTIONS_PATH}/${serviceInstance.slug}/${document.slug}`;
+
+  const fallbackDescription =
```
Comment (Kroustille) [inline apps/portal-front/app/[locale]/(public)/cybersecurity-solutions/[slug]/[docSlug]/page.tsx]: The service could also be OpenAEV scenario. Isn't it necessary to also have a fallback description for this case ?
Or maybe we should have a generic fallback description in the case we add other services.

Kroustille (commented):
Since the locale part of URL is only used on public paths, I would have put `[locale]` folder inside `(public)` instead of putting `(public)` inside `[locale]`.

This way, we would keep an easy view of which folder corresponds to what at the root of `/app`. If it's planned to add locale on another path, we should keep it how it's currently done 😉 

What do you think ?

apps/portal-front/app/[locale]/(public)/cybersecurity-solutions/[slug]/[docSlug]/page.tsx
Comment (Julienger0) [reply apps/portal-front/app/[locale]/(public)/cybersecurity-solutions/[slug]/[docSlug]/page.tsx]: I added the fallback description + you were totally right about putting [locale] folder inside (public)

apps/frontend/app/(public)/[locale]/layout.tsx
```diff
@@ -1,18 +1,43 @@
...
+import { publicLocales, type PublicLocale } from '@/i18n/config';
 import { getDefaultMetadata } from '@/utils/generate-metadata';
 import { Button } from '@filigran/ui/servers';
 import '@filigran/ui/theme.css';
 import LogoXTMDark from '@public/logo_xtm_hub_dark.svg';
 import '@styles/globals.css';
 import { Metadata } from 'next';
-import Link from 'next/link';
+import { getTranslations, setRequestLocale } from 'next-intl/server';
+import { default as Link, default as NextLink } from 'next/link';
```
Comment (jpkha) [inline apps/frontend/app/(public)/[locale]/layout.tsx]: Isn't it the same ? Link and NextLink

apps/frontend/src/i18n/navigation.ts
```diff
@@ -0,0 +1,9 @@
+import { defaultLocale, locales } from '@/i18n/config';
+import { createNavigation } from 'next-intl/navigation';
+
+export const { Link, redirect, usePathname, useRouter, getPathname } =
```
jpkha: Where do you use this files ? I don't find any reference ?

jpkha:
IMO when you access to an old url like "https://hub.staging.filigran.io/cybersecurity-solutions/openaev-scenarios" you should be redirected or get access to the default language which is english.

apps/frontend/app/(public)/[locale]/layout.tsx
Comment (Julienger0) [reply apps/frontend/app/(public)/[locale]/layout.tsx]: Actually it's a leftover from my initial modifications. It's no longer needed after the changes, I just missed it

apps/frontend/src/i18n/navigation.ts
Julienger0: same thing here

jpkha:
Missing traduction in the breadcrumb : 
image

jpkha:
I cannot access to any opencti-integrations details:
 https://dev-pr-2238.hub.staging.filigran.io/en/cybersecurity-solutions/opencti-integrations 
 Also the share link have the language path (the button in the card).

apps/e2e/tests/tests_files/xtm-platform-roadmap.spec.ts
```diff
@@ -122,7 +122,7 @@ test.describe('XTM Platform Roadmap', () => {
       uploader_id: ADMIN_USER.ID,
     });
 
-    await page.goto('/cybersecurity-solutions/xtm-platform-roadmap');
+    await page.goto('/en/cybersecurity-solutions/xtm-platform-roadmap');
```
jpkha: If the URL is automatically redirected to /en/, do we really need to force /en/ in the URL? I’d say we should keep the previous behavior and let the test cover the redirection instead.

apps/frontend/src/utils/services.ts
```diff
@@ -228,13 +247,15 @@ export const publicServiceInstanceToInstanceCardData = (
 };
 
 export const seoServiceInstanceToInstanceCardData = (
-  instance: seoServiceInstanceFragment$data
+  instance: seoServiceInstanceFragment$data,
+  t: ReturnType<typeof useTranslations>,
+  locale?: string
```
jpkha: I don't think you need to pass the locale anymore

jpkha:
Missing code for sitemap, but looks like is not working anymore, in Prod the sitemap is down :
https://hub.filigran.io/sitemap.xml

jpkha (approved):

apps/frontend/proxy.ts
```diff
@@ -1,12 +1,33 @@
...
   const proxyResponse = await manageRequest(request);
   if (proxyResponse) {
     return proxyResponse;
   }
 
+  const { pathname } = request.nextUrl;
+
+  if (
+    PUBLIC_LOCALE_PATH.test(pathname) ||
+    pathname.startsWith('/cybersecurity-solutions')
```
jpkha: I think it would be better to approach this with `!pathname.startsWith('/app')` We are sure that the app is the private parts.
IMO for public path could be different than cybersecurity-solutions

Issue #2255, opened 2026-04-29
[tech] remove extractId for different modules


carinelebas:
Remove extractId for:
- document
- user
- use case
- subscription

PR #2256, opened 2026-04-29
[backend/frontend] chore(extractId): remove extract id for different modules and use scalar (#2255)

carinelebas:
# Context
Remove extractId for:
- document
- user
- use case
- subscription

## How to test
Test there is no regression on these modules

## Related issues

- Related to #2255

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/portal-api/src/modules/deployment/group/service-group.graphql
```diff
@@ -8,7 +8,7 @@ type ServiceGroup implements Node {
 
 input UpdateServiceGroupsInputGroup {
   id: ServiceGroupId!
-  userIds: [ID!]!
```
hervyt, line 11: **thought** : 
I'm currently the head inside ESLint but.... What about a new rule to forbid ID type ? 🗡️  😄

carinelebas: I think it's a bit too restrictive 😄  It's needed at least for all types implementing Node, and it's also used in pagination, where having a specific type could make the code more complex.

Kroustille (approved):

Issue #2261, opened 2026-04-30
Add a tag EE or partial EE to differentiate EPICS


EllynBsc:
**Overall Description**
Add a tag:
- EE 
- partial CE/EE

to differentiate EPICS

PR #2300, opened 2026-05-11
[backend/frontend] feat(Roadmap): Add tag EE (#2261)

hervyt:
# Context
> This PR adds a tag EE/Partial EE if needed. It adds the question in the form. 
> It also changes the wording of the slack sentence
> It also removes XTM One from the filter

## How to test
- Log in as user with the irghts to add/update epics. 
- add a new epic. By default, it's CE and should not add the tag on the epic card. 
- Add a new epic with EE. The tag EE should be displayed on the card. 
- Add a new epic with Partial EE, the tag Partial EE should be displayed on the card. 
- Update an epic EE to CE : the tag should not be displayed anymore. 
- Update an epic CE to EE : the tag should be displayed. 

## Related issues

- Related to #2261 

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [ ] Integration tests
- [X] E2E tests
- [X] Local manual tests


apps/backend/src/migrations/20260511100000_add_edition_type_to_epic.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  return knex.schema.alterTable('Epic', function (table) {
+    table
+      .enum('edition_type', [
+        'enterprise_edition',
+        'partial_ee',
```
Kroustille, line 10: Since the other values are fully written, I would say we can use `partial_enterprise_edition` here. What do you think ?

apps/frontend/src/components/epic/epic-item/EditionTypeMapping.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { EditionTypeEnum } from '@generated/models/EditionType.enum';
+import { describe, expect, it } from 'vitest';
+import { EditionTypeMapping } from './EditionTypeMapping';
+
+describe('EditionTypeMapping', () => {
+  it('contains all Edition Type items from enum', () => {
```
Kroustille: This test is not necessary because `EditionTypeMapping` is a record with `EditionTypeEnum` as key and is not partial: you should have a typescript error when an item is missing

Kroustille (approved):

apps/backend/src/migrations/20260511100000_add_edition_type_to_epic.js
hervyt: I thought about it, however, this variable name is particularly long, and does not improve the code, as we understand what this is...

PR #2343, opened 2026-05-19
[frontend] feat(codegen/react-query): implement new stack (#2355)

jpkha:
# GraphQL Codegen Cookbook (Frontend)

This guide explains, step by step, how to add or update a GraphQL operation and verify generated `use...` hooks and `mock...` factories.

---

## 1) Run in dev mode

- Yarn install
- Launch dev cmd

---

## 2) Where to write GraphQL operations

GraphQL documents are located in:

- `apps/frontend/graphql/**/*.graphql`

Existing examples:

- `apps/frontend/graphql/use-case/use-case.query.graphql`
- `apps/frontend/graphql/use-case/use-case.mutation.graphql`

---

### Query example

```graphql
query UseCasesList($count: Int!, $orderBy: UseCaseOrdering!, $orderMode: OrderingMode!) {
  useCases(first: $count, orderBy: $orderBy, orderMode: $orderMode) {
    totalCount
    edges {
      node {
        id
        name
        color
      }
    }
  }
}
```

### Mutation example

```graphql
mutation UseCaseAdd($input: AddUseCaseInput!) {
  addUseCase(input: $input) {
    id
    name
    color
  }
}
```

> Important: the operation name (`UseCasesList`, `UseCaseAdd`, etc.) is used for generated names and MSW mocks through `queryName`.

---

## 4) Run generation

Generation should be automatic

Expected outputs:

- `apps/frontend/graphql/generated.ts`
- `apps/frontend/graphql/mocks.ts`

---

### 5.1 Hooks/types (`generated.ts`)

Look for:

- operation types: `UseCaseAddMutation`, `UseCasesListQuery`, etc.
- React Query hooks: `useUseCaseAddMutation`, `useUseCasesListQuery`, etc.

### 5.2 Mocks (`mocks.ts`)

Look for `mock...` factories, for example:

- `mockUseCase(...)`
- `mockUseCaseEdge(...)`
- `mockUseCaseConnection(...)`

---

## 6) Use in components

Query example:

```ts
const { data } = useUseCasesListQuery(portalGraphqlClient, {
  count: 100,
  orderMode: OrderingMode.Asc,
  orderBy: UseCaseOrdering.Name,
});
```

Mutation example:

```ts
const { mutate: createUseCase } = useUseCaseAddMutation(portalGraphqlClient);
createUseCase({ input: { name: 'Threat hunting', color: '#11aa22' } });
```

---

### Query mock

```ts
mswServer.use(
  mockGraphqlQuery({
    queryName: 'UseCasesList',
    data: {
      useCases: mockUseCaseConnection({
        totalCount: 1,
        edges: [
          mockUseCaseEdge({
            node: mockUseCase({ id: 'id-1', name: 'Threat hunting', color: '#11aa22' }),
          }),
        ],
      }),
    },
  })
);
```

### Mutation mock (success)

```ts
mswServer.use(
  mockGraphqlMutation({
    queryName: 'UseCaseAdd',
    data: {
      addUseCase: mockUseCase({ id: 'id-1', name: 'Threat hunting', color: '#11aa22' }),
    },
  })
);
```

### Mutation mock (error)

```ts
mswServer.use(
  mockGraphqlMutation({
    queryName: 'UseCaseAdd',
    errors: [{ message: 'InternalError' }],
  })
);
```

---
If needed, use an existing test in `apps/frontend/src/components/admin/use-case/` as a reference.


apps/frontend/src/lib/graphql-client.ts
```diff
@@ -0,0 +1,25 @@
...
+    ? prepareUri(process.env.SERVER_HTTP_API)
+    : prepareUri(process.env.NEXT_PUBLIC_CLIENT_HTTP_API);
+
+  return baseUri + 'graphql-api';
+}
+
+const graphqlApiEndpoint = resolveGraphqlApiEndpoint();
+
+export const portalGraphqlClient = new GraphQLClient(graphqlApiEndpoint, {
+  credentials: 'same-origin',
```
carinelebas: just wondering: are there any other parameters we should add here for security reasons ?

carinelebas (approved):

apps/frontend/graphql/use-case/use-case.query.graphql
```diff
@@ -0,0 +1,22 @@
...
+        name
+        color
+      }
+    }
+  }
+}
+
+
+
+
```
Kroustille: This trailing lines makes me think we don't have a linter/formatter for graphql files in frontend

apps/frontend/scripts/postprocess-codegen.mjs
```diff
@@ -0,0 +1,16 @@
...
+const generatedPath = path.resolve('graphql/generated.ts');
+
+if (!fs.existsSync(generatedPath)) {
+  process.exit(0);
+}
+
+const lines = fs.readFileSync(generatedPath, 'utf8').split('\n');
+
+lines[0] = 'import type { GraphQLClient, RequestOptions } from "graphql-request";';
+lines[1] = 'type GraphQLClientRequestHeaders = RequestOptions["requestHeaders"];';
```
Kroustille: Is it intended to replace content of lines 0 and 1 with this content instead of using unshift ? 🤔

apps/frontend/src/components/admin/use-case/EditUseCase.tsx
```diff
@@ -1,31 +1,85 @@
...
+  const { mutate: editUseCase } = useUseCaseEditMutation(portalGraphqlClient, {
+    onSuccess: () => {
+      toast({
+        title: t('Utils.Success'),
+      });
+      queryClient.invalidateQueries({
+        queryKey: useUseCasesListQuery.getKey({
+          count: 100,
+          orderMode: OrderingMode.Asc,
+          orderBy: UseCaseOrdering.Name,
```
Kroustille, line 40: I think all mutations/queries needs to get their key from a common place. Otherwise it can leads to errors if we update the query key parameters and not the mutation ones

apps/frontend/graphql/use-case/use-case.query.graphql
```diff
@@ -0,0 +1,22 @@
+fragment UseCaseRow on UseCase {
+  id
+  name
+  color
+}
```
Kroustille, line 5: Can we use this fragment inside the `useCases` query below ?

apps/frontend/src/lib/graphql-client.ts
jpkha: I will check, thanks

apps/frontend/graphql/use-case/use-case.query.graphql
jpkha: yes we can. Actually we are not using the response for the moment

apps/frontend/graphql/use-case/use-case.query.graphql
jpkha: yes it's running but I don't know why the hook is not working well. Will do in a next PR

apps/frontend/scripts/postprocess-codegen.mjs
jpkha: yes it's intended to replace the content. As it's generated we know what happened. Issue with the graphql-request v6 import but will not fixe it I think.
For the moment this is the best solution we have.
Unshift would only remove the line but we need to replace it with this specific content.

apps/frontend/src/lib/graphql-client.ts
jpkha: In my opinion, the security aspect should mainly be handled server-side, since user input should never be fully trusted.

If we want to add an extra layer of security or better identify the user, we could include additional information from the headers.

apps/frontend/graphql/use-case/use-case.query.graphql
jpkha: done

apps/frontend/src/components/admin/use-case/EditUseCase.tsx
jpkha: https://github.com/FiligranHQ/xtm-hub/issues/2374

PR #2344, opened 2026-05-19
[backend/frontend] feat(trials): adapt use cases to openaev trials (#2156)

carinelebas:
# Context
We need to have different use cases for openAEV trials vs openCTI trials.

## How to test
- Start an openAEV trial from banner and from details page: it should display the openAEV use cases
- Start an openCTI trial from banner and from details page: it should display the openCTI use cases

## Related issues

- Related to #2156

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/src/components/service/trial-instances/form-constants.ts
```diff
@@ -9,3 +11,19 @@ export const REGIONS = Object.values(DeploymentRequestPlatformRegionEnum).map(
...
+  [PlatformIdentifierEnum.OPENAEV]: (
+    Object.values(
+      DeploymentRequestUseCaseEnum
+    ) as DeploymentRequestUseCaseEnum[]
+  ).filter((v) => v.startsWith('oaev_')),
+  [PlatformIdentifierEnum.OPENCTI]: (
+    Object.values(
+      DeploymentRequestUseCaseEnum
+    ) as DeploymentRequestUseCaseEnum[]
+  ).filter((v) => !v.startsWith('oaev_')),
```
Kroustille, line 28: Shouldn't we also prepend deployment request use case with octi and filter it there ? Otherwise we will have an issue when there is more than 2 products

apps/frontend/src/components/service/trial-instances/form-constants.ts
```diff
@@ -9,3 +11,19 @@ export const REGIONS = Object.values(DeploymentRequestPlatformRegionEnum).map(
     label: region.toUpperCase(),
   })
 );
+
+export const USE_CASES_BY_PLATFORMIDENTIFIER: Record<
```
Kroustille:
```suggestion
export const USE_CASES_BY_PLATFORM_IDENTIFIER: Record<
```

I think it's easier to read, do you agree ?

Kroustille (approved):

apps/frontend/src/components/service/trial-instances/form-constants.ts
carinelebas: It's not so easy to change this value as it's not internal but also send in telemetry, so changing it requires coordination with other teams. I agree it would be nice to align, but I'm not sure it's worth the trouble.

apps/frontend/src/components/service/trial-instances/form-constants.ts
carinelebas: done in [5188dca](https://github.com/FiligranHQ/xtm-hub/pull/2344/commits/5188dcae350294cf08c9b18f472c3a66fd522095)

PR #2362, opened 2026-05-21
[frontend/backend] refactor(Subscription): Rework userService management (#2160)

hervyt:
# Context
> Improve the subscription's details page with a list of userServices. Bulk selection to add capabilities or delete accesses. Rework the userService edition. 

## How to test
- Connect as ADMIN_BYPASS 
- Go to the service list 
- Choose a service, subscribe an organization to this service and then click "manage users". 
- You can invite multiple users in a bulk in this service. You can bulk delete them. You can bulk add capabilities to them. You can edit one by one if you prefer. 

## Related issues

- Related to #2160

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [X] Integration tests
- [X] E2E tests
- [X] Local manual tests


apps/frontend/src/components/subcription/[slug]/SubscriptionSlug.tsx
```diff
@@ -151,25 +155,28 @@ const SubscriptionSlug = ({
...
-        },
-        onCompleted() {},
-      });
-    },
-    [
-      commitUserServiceDeletingMutation,
-      userServices.userServiceFromSubscription?.__id,
-      subscriptionId,
-    ]
+  const isBypass = useMemo(
```
Comment (carinelebas) [inline apps/frontend/src/components/subcription/[slug]/SubscriptionSlug.tsx]: maybe we could use `useAdminByPass()` here?

apps/backend/src/modules/user-service/user-service.domain.ts
```diff
@@ -197,6 +199,10 @@ export const UserServiceDomain = {
     return db<UserService>('User_Service').where(field);
   },
 
+  loadUserServicesBy: async (ids: UserServiceId[]): Promise<UserService[]> => {
```
carinelebas: This function should rather be named `loadUserServicesByIds`, function ending simply by `by` usually accept a mutator in parameter.

apps/backend/src/modules/security-management/user-service-capability/user-service-capability.helper.ts
```diff
@@ -53,61 +55,79 @@ export const insertCapabilities = async (
...
+    ? await db<UserServiceCapability>('UserService_Capability')
+        .whereIn('user_service_id', userServiceIds)
+        .select(
+          'user_service_id',
+          'generic_service_capability_id',
+          'subscription_capability_id'
+        )
+    : [];
 
-  const [insertedCapability] = await db<UserServiceCapability>(
```
carinelebas: why didn't you use this query ? it looks to do the same in a simpler way?

apps/backend/src/modules/security-management/subscription-capability/subscription-capability.domain.ts
```diff
@@ -29,7 +30,8 @@ export const addCapabilitiesToSubscriptions = async (
 
   const data = subscriptionIds.flatMap((subscriptionId) =>
     capabilityIds.map((capabilityId) => ({
-      service_capability_id: capabilityId,
+      service_capability_id: fromGlobalId(capabilityId)
```
carinelebas: question: can't we use a scalar for this? otherwise extractId would be better.

apps/backend/src/modules/security-management/user-service-capability/user-service-capability.app.ts
```diff
@@ -1,17 +1,29 @@
...
   addCapabilitiesToUserServices: async (
     userServiceIds: UserServiceId[],
-    capabilities: string[]
+    capabilities: string[],
+    serviceInstanceId: ServiceInstanceId
   ): Promise<UserService[]> => {
     const userServices =
       await UserServiceDomain.loadUserServicesBy(userServiceIds);
 
+    const isUserServiceInCorrectService =
```
carinelebas, line 19: maybe this could be moved to security guard?

apps/backend/src/modules/subscription/subscription.app.test.ts
```diff
@@ -110,8 +114,12 @@ describe('subscription app', () => {
         end_date: endDate,
       });
       await addCapabilitiesToSubscription(id, [
-        SERVICES.INSTANCES.INTEGRATIONS.CAPABILITIES.UPLOAD.ID,
-        SERVICES.INSTANCES.INTEGRATIONS.CAPABILITIES.DELETE.ID,
+        toCapabilityGlobalId(
+          SERVICES.INSTANCES.INTEGRATIONS.CAPABILITIES.UPLOAD.ID
```
carinelebas: maybe it would be easier to have a `SERVICES.INSTANCES.INTEGRATIONS.CAPABILITIES.UPLOAD.GLOBAL_ID` defined? (if really needed)

apps/backend/src/modules/security-management/subscription-capability/subscription-capability.domain.ts
carinelebas: Also, are you fixing a bug? Since you are also updating all test cases, it seems you are changing the behavior of the function. Are there no side effect on another call path? 
Anyway, if a conversion is needed, I think it would be easier to maintain if the conversion was done in the resolver rather than in the domain.

apps/backend/src/modules/user-service/user-service.app.ts
```diff
@@ -35,29 +37,33 @@ export const UserServiceApp = {
...
+    const userService =
+      await UserServiceDomain.loadUserServiceById(userServiceId);
+    // Check the provided subscription_id correspond to the correct userService
+    if (userService.subscription_id !== subscriptionId) {
+      throw new Error(ForbiddenErrorCode.ServiceNotManageable);
     }
-
-    return deletedUserService;
+    await UserServiceDomain.deleteUserCapabilityById(userService.id);
+    await insertCapabilities(capabilities, [userService]);
```
carinelebas: should delete and insert be in a transaction ?

apps/backend/src/modules/user-service/user-service.graphql
```diff
@@ -32,23 +31,33 @@ input UserServiceEditInput {
...
     @auth(
       portalCapa: [BYPASS]
       orgaCapa: [ADMINISTRATE_ORGANIZATION, MANAGE_SUBSCRIPTION]
     )
-  @service_capa(requires: [MANAGE_ACCESS])
-
+    @service_capa(requires: [MANAGE_ACCESS])
+  editUserService(
+    input: UserServiceEditInput!
+    subscription_id: SubscriptionId
```
carinelebas: just some thinking: I understand that the directive needs direct access to the subscription_id (or service_instance_id), but I'm not sure it's the good move to extract it from the input.

Some explanation of my concern by AI (better formulated than me 😄 ): 
Extracting subscription_id from the input solely to satisfy the directive's technical requirements leaks an implementation detail into the public API contract. The input type has semantic cohesion — its fields belong together because they describe the operation, not because of how the security layer is implemented. Promoting one field to a top-level argument for a non-functional reason breaks that cohesion and signals something to API consumers that isn't true: that subscription_id has a different status or role than the rest of the input. The schema should reflect the domain, not the constraints of the authorization infrastructure.

apps/backend/src/modules/security-management/subscription-capability/subscription-capability.domain.ts
hervyt: The behaviour is changing yes as new features are implemented :) Now everything can be bulk selected, so everything is transformed into a tab ;)

apps/backend/src/modules/subscription/subscription.app.test.ts
hervyt: not needed now with scalar :)

apps/frontend/src/components/service/[slug]/UserServiceForm.tsx
```diff
@@ -187,8 +190,10 @@ export const UserServiceForm = ({
         input: {
           email: values.email.map(({ text }) => text),
           capabilities: values.capabilities,
-          subscriptionId: subscription.subscriptionById?.id ?? '',
+          subscription_id: subscription.subscriptionById!.id,
```
Comment (carinelebas) [inline apps/frontend/src/components/service/[slug]/UserServiceForm.tsx]: this can fail at runtime, is there not a better way to ensure it's not null? 
it's forbidden on the backend, should it not also be on the frontend?

carinelebas (approved):

Issue #2371, opened 2026-05-22
Store last connectivity check date in DB


carinelebas:
## Description

To better know which platforms are really active, we need to store the last connectivity check date in DB.

PR #2411, opened 2026-05-28
[frontend/backend] feat(Registration): get last_connectivity_check and status on frontend (#2371)

hervyt:
# Context
> Update the service contract to get the last_connection_check in DB. On every connection_check, we must update this date. 
> On the frontend, on the detailed page, we should display the "last_connection_date" and display "connected/disconnected". 

## How to test
- Register an OpenCTI instance on XTM Hub. 
- You should see the date on last connectivity_check 
- Wait a few days : this date should be updated. 
- CHeck there is no regression on trials. 
- Check a trial can still auto register :) 

## Related issues

- Related to #2371

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [X] Local manual tests

# Additional information


apps/backend/src/migrations/20260527120000_add-connection-fields-to-service-contract-schema.js
```diff
@@ -0,0 +1,71 @@
+const LAST_CONNECTION_CHECK_PROPERTY = {
+  type: 'string',
+  format: 'string',
```
carinelebas: question: can't we use a date type rather than a string? it's usually more efficient and clearer to store dates?

apps/frontend/src/components/service/registration/RegistrationDetails.tsx
```diff
@@ -87,18 +87,27 @@ export const RegistrationDetails = ({
     platform.deployment_request?.hub_status ===
       DeploymentRequestHubStatusEnum.ACTIVE;
 
+  const isConnectionStatusOk = isWithinLastMonths(
```
carinelebas, line 90: should it be `isWithinLastMonth` if only one month?

hervyt: It's a generic function where the number of months is a parameter... It's often used with "3" months so I'm not sure to rename it ? :thinking:

carinelebas: oops..looks like I'm still sleeping :)

carinelebas (approved):

apps/backend/src/migrations/20260527120000_add-connection-fields-to-service-contract-schema.js
```diff
@@ -0,0 +1,71 @@
...
+    schema?.properties && typeof schema.properties === 'object'
+      ? schema.properties
+      : {};
+
+  return {
+    ...schema,
+    required: [...schema.required, 'last_connection_check'],
+    properties: {
+      ...properties,
+      last_connection_check: LAST_CONNECTION_CHECK_PROPERTY,
```
Kroustille: It would also needs to be updated in the whole PR codebase and would be more coherent with the connectivity app and enum.
```suggestion
      last_connectivity_check: LAST_CONNECTIVITY_CHECK_PROPERTY,
```

Kroustille (changes requested):
I propose to rename every connection to connectivity to fit with current naming. What do you think ? 🙂

Kroustille (approved):

apps/backend/src/modules/registration/registration.app.ts
```diff
@@ -512,6 +513,8 @@ const mapDomainRegisteredPlatformToGraphQL = (
     __typename: 'RegisteredPlatform',
     id: platform.id,
     platform_id: platform.config?.platform_id ?? platform.id,
+    last_connectivity_check:
+      platform.config?.last_connectivity_check ?? new Date(),
```
Kroustille: I think we should return null when there is no last connectivity check, meaning the platform is not active anymore

Issue #2442, opened 2026-06-02
[backend] Fix typescript strictNullChecks - part 1


carinelebas:
First pass of strictNullChecks:
- make some db fields non-nullable
- make some graphql parameters mandatory
- fix return types in queries with returning
- properly type db() calls
- transform inputMaybe null to undefined in inputs

PR #2443, opened 2026-06-02
[backend] chore: fix typescript strictNullChecks - part 1 (#2442)

carinelebas:
# Context
First pass of strictNullChecks fixes:

- make some db fields non-nullable
- make some graphql parameters mandatory
- fix return types in queries with returning
- properly type db() calls
- transform inputMaybe null to undefined in inputs

test there is no regression

## Related issues

- Related to #2442

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

### DIFF
--- a/apps/backend/knexfile.ts
+++ b/apps/backend/knexfile.ts
@@ -551,7 +551,7 @@ export const paginate = async <T extends object, U>(
   applyLogicalFilter(type, queryContext, logicalFilters ?? undefined);
 
   if (filters) {
-    await applyFilters(type, queryContext, filters ?? undefined);
+    await applyFilters(type, queryContext, filters);
   }
 
   await applySearch(
--- a/apps/backend/knexfile.ts
+++ b/apps/backend/knexfile.ts
@@ -92,13 +92,13 @@ export type ActionType = 'add' | 'edit' | 'delete' | 'merge' | 'invalidate';
 export type MethodType = 'select' | 'insert' | 'update' | 'del';
 
 interface Pagination {
-  first?: number;
-  after?: string;
-  orderMode?: string;
-  orderBy?: string;
-  filters?: Filters;
-  logicalFilters?: LogicalFilterInput;
-  searchTerm?: string;
+  first?: number | null;
+  after?: string | null;
+  orderMode?: string | null;
+  orderBy?: string | null;
+  filters?: Filters | null;
+  logicalFilters?: LogicalFilterInput | null;
+  searchTerm?: string | null;
 }
 
 const knex = pkg;
@@ -139,23 +139,21 @@ export const database = knex(config);
 
 export interface QueryOpts {
   unsecured?: boolean;
-  first?: number;
-  after?: string;
-  orderMode?: string;
-  orderBy?: string;
+  first?: number | null;
+  after?: string | null;
+  orderMode?: string | null;
+  orderBy?: string | null;
   methodType?: MethodType;
   capabilities?: string[];
-  searchTerm?: string;
-  filters?: Filters;
+  searchTerm?: string | null;
+  filters?: Filters | null;
   columns?: string[];
   normalizeSearchTerm?: boolean;
 }
 
 export const dbRaw: Knex.RawBuilder = database.raw.bind(database);
 
-export const dbTx = () => database.transaction();
-
-export function db<T>(
+export function db<T extends object>(
   type: DatabaseType
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): Knex.QueryBuilder<T, any> {
@@ -164,7 +162,8 @@ export function db<T>(
   });
 
   if (databaseContext.isInTransaction()) {
-    queryContext.transacting(databaseContext.getTransaction());
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    queryContext.transacting(databaseContext.getTransaction()!);
   }
 
   return queryContext;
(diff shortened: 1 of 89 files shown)


apps/backend/src/modules/organization-management/user/user-organization/user-organization.domain.ts
```diff
@@ -73,16 +76,17 @@ export const UserOrganizationDomain = {
 
   updateMultipleUserOrgWithCapabilities: async (
     userId: UserId,
-    orgCapabilities?: OrganizationCapabilitiesInput[]
+    orgCapabilities?: OrganizationCapabilitiesInput[] | null
```
Kroustille: A default value can be used to not declare another `safeCapabilities` constant below.

```suggestion
    orgCapabilities: OrganizationCapabilitiesInput[] | null = []
```

apps/backend/src/modules/security-management/user-organization-capability/user-organization-capability.domain.ts
```diff
@@ -7,42 +7,50 @@ import UserOrganizationCapability, {
 
 export const createUserOrganizationCapability = async ({
   user_organization_id,
-  capabilities_name = [],
+  capabilities_name,
 }: {
   user_organization_id: UserOrganizationId;
-  capabilities_name: string[];
+  capabilities_name?: string[] | null;
```
Kroustille: It could be possible to use default value by using this suggestion I think.
```suggestion
  capabilities_name: string[] | null | undefined;
```

It's easier to read than using a parameter and than a default value in a second constant. What do you think ?

apps/backend/src/modules/security-management/user-organization-capability/user-organization-capability.domain.ts
```diff
@@ -7,42 +7,50 @@ import UserOrganizationCapability, {
...
 };
 
 export const updateUserOrganizationCapability = async ({
   user_organization_id,
-  capabilities_name = [],
+  capabilities_name,
 }: {
   user_organization_id: UserOrganizationId;
-  capabilities_name: string[];
+  capabilities_name?: string[] | null;
```
Kroustille: Same here for the default value in parameter 😄

apps/backend/src/modules/service/definition/service-definition.domain.ts
```diff
@@ -50,7 +53,7 @@ export const ServiceDefinitionDomain = {
       serviceDefinitionIdentifierMappedByPlatformIdentifier[platformIdentifier];
 
     if (!serviceDefinitionIdentifier) {
-      return undefined;
+      return Promise.resolve(undefined);
```
Kroustille: Instead of a returning a `Promise.resolve`, it's also possible to mark the function as `async`.

apps/backend/src/modules/shareable-resource/opencti/integration/ingest-manifest/ingest-manifest.graphql
```diff
@@ -1,4 +1,4 @@
 type Query {
-  updateOpenCTIManifest(tag: String): Success!
+  updateOpenCTIManifest(tag: String!): Success!
```
Kroustille, line 2: We need to be careful because this query is called from the n8n automation to ingest manifests, and it needs to have the tag parameter as required too. 
Otherwise the automation will fail with a GraphQL validation error.

apps/frontend/app/(application)/app/page.tsx
```diff
@@ -54,13 +54,14 @@ const Page = () => {
     useQueryLoader<registerRegisteredPlatformsQuery>(
       RegisterRegisteredPlatformsQueryGraphql
     );
-  useMountingLoader(loadQueryRegisteredPlatforms, {
-    fetchPolicy: 'network-only',
-  });
```
Kroustille, line 59: Is it normal the `fetchPolicy` disappeared ?

apps/backend/knexfile.ts
```diff
@@ -549,11 +549,18 @@ export const paginate = async <T, U>(
     connection: true,
   });
 
-  applyLogicalFilter(type, queryContext, logicalFilters);
+  applyLogicalFilter(type, queryContext, logicalFilters ?? undefined);
 
-  await applyFilters(type, queryContext, filters);
+  if (filters) {
+    await applyFilters(type, queryContext, filters ?? undefined);
```
Kroustille:
```suggestion
    await applyFilters(type, queryContext, filters);
```

`?? undefined` should be necessary because the `applyFilters` call is in a condition where we check if filters is not null nor undefined 🤔

apps/backend/src/utils/error/error.code.ts
```diff
@@ -102,6 +102,7 @@ export enum NotFoundErrorCode {
   DeploymentRequestQuotaNotFound = 'DEPLOYMENT_REQUEST_QUOTA_NOT_FOUND',
   DocumentNotFound = 'DOCUMENT_NOT_FOUND',
   NewsFeedItemNotFound = 'NEWS_FEED_ITEM_NOT_FOUND',
+  UserNotFound = 'USER_NOT_FOUND',
```
Kroustille, line 104: This new error is not translated.

apps/backend/src/modules/shareable-resource/opencti/integration/ingest-manifest/ingest-manifest.graphql
carinelebas: I already checked n8n, it's only making a graphql call, and we should always have the tag, so there is nothing to change there.

apps/frontend/app/(application)/app/page.tsx
carinelebas: yes, it was not the right place :) the second parameter is the variables, not the options. Options should not be added in useMountingLoader, but when the function is called.

Kroustille (approved):

Issue #2374, opened 2026-05-22
Improve/Propose QueryKey management with React-Query


jpkha:
Create a "centralized" source for React Query keys used by both queries and mutations. This will help prevent inconsistencies and potential bugs when query key parameters are updated in one place but not reflected in related mutations.

PR #2448, opened 2026-06-03
[frontend] refactor(react-query): refactor code and add store key (#2374)

jpkha:
# Query Key Usage (Use Cases)

This document describes how query keys are used for the Use Case list and how cache is updated after mutations.

## 1) Query key factory

Source of truth: `apps/frontend/graphql/use-case/use-case-list.keys.ts`

```ts
export const useCaseListKeys = {
  all: useUseCasesListQuery.getRootKey,
  list: (variables: UseCasesListQueryVariables) =>
    useUseCasesListQuery.getKey(variables),
};
```

- `all()` targets the root list key (`['UseCasesList']`).
- `list(variables)` returns the full generated key for a specific variable set.

## 2) Reading data with the shared key

In `UseCases.tsx`, the list query uses the same root key:

```ts
useUseCasesListQuery(portalGraphqlClient, useCasesListVariables, {
  queryKey: useCaseListKeys.all(),
});
```

This ensures every cache update using `useCaseListKeys.all()` applies to the displayed list.

## 3) Updating cache after mutations (without invalidation)

Instead of calling `invalidateQueries`, mutations update cache directly with `setQueryData`.

### Add

In `AddUseCase.tsx`:

```ts
queryClient.setQueryData<UseCasesListQuery>(
  useCaseListKeys.all(),
  prependToQueryCache('useCases', { node: newUseCase })
);
```

Effect:
- inserts the new edge at the top,
- increments `totalCount`.

### Edit

In `EditUseCase.tsx`:

```ts
queryClient.setQueryData<UseCasesListQuery>(
  useCaseListKeys.all(),
  updateInQueryCache('useCases', data.editUseCase)
);
```

Effect:
- replaces the matching node by `id` in `edges`.

### Delete

In `EditUseCase.tsx`:

```ts
queryClient.setQueryData<UseCasesListQuery>(
  useCaseListKeys.all(),
  removeFromQueryCache('useCases', data.deleteUseCase.id)
);
```

Effect:
- removes the edge with the matching `id`,
- decrements `totalCount` (never below `0`).

## 4) Generic cache helpers

Helpers are in `apps/frontend/src/utils/query-cache.ts`:

- `prependToQueryCache(connectionKey, newEdge)`
- `updateInQueryCache(connectionKey, updatedNode)`
- `removeFromQueryCache(connectionKey, deletedId)`

All helpers are pure updater functions, designed to be passed directly to `setQueryData`.

## 5) Recommended pattern

1. Define keys in one factory file.
2. Use the same key in the query and mutation cache updates.
3. Prefer direct cache updates (`setQueryData`) for add/edit/delete on connection lists.
4. Use `list(variables)` when you need a variable-specific key; use `all()` for global list targeting.


apps/frontend/graphql/use-case/use-case-list.keys.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  type UseCasesListQueryVariables,
+  useUseCasesListQuery,
+} from '@graphql/generated';
+
+export const useCaseListKeys = {
+  all: () => ['UseCasesList'] as const, // the key is automatically generated by codegen
```
Kroustille: Does it work with:



to prevent having a magic string here and directly use the one generated ? 🤔 
```suggestion
  all: () => useUseCasesListQuery.getKey() as const, // the key is automatically generated by codegen
```

apps/frontend/src/components/admin/use-case/EditUseCase.tsx
```diff
@@ -29,17 +31,14 @@ const EditUseCase = ({
...
-      queryClient.invalidateQueries({
-        queryKey: useUseCasesListQuery.getKey({
-          count: 100,
-          orderMode: OrderingMode.Asc,
-          orderBy: UseCaseOrdering.Name,
-        }),
-      });
+      queryClient.setQueryData<UseCasesListQuery>(
+        useCaseListKeys.all(),
+        updateInQueryCache('useCases', data.editUseCase)
```
Kroustille, line 40: It would be nice to declare the `useCases` query cache key somewhere instead of using magic strings

jpkha: the useCases is not the query cache key but I can try to check how to handle without using a "magic strings"

jpkha: add a "& keyof Query" to check if the key still exist

Kroustille (approved):

PR #2462, opened 2026-06-03
chore(ci): Format backend frontend e2e and add i18n check (#2463)

jpkha:
- Run in CI Format check for e2e, front, backend.
- Fix format issue on the repo
- Add e2e lint and prettier command and config
- Add check-i18n format, will check if the variable in each language is correct and missing key
- Add new eslint rule format-js to detect missing i18n variable, add missing one accordingly

In case of i18n error exemple  :
image


hervyt (approved):
💯

Issue #2449, opened 2026-06-03
fix(playbooks): cannot add certain people to service playbooks


elisavaldegas:
**Describe the bug**
Not possible to see the full list of filigran users please add them we only have 50 at the moment

PR #2473, opened 2026-06-04
feat(adminservice): fix Multiselect to refresh (#2449)

hervyt:
# Context
> Before the option, the multiselect did not refreshed and when we searched for people that are not in the first page of pagination > We couldn't find them. 

## How to test
- Add 51 organizations (the last with a z at the beginning of the name. 
- Click on subscribe an organization on the subscription page. 
- In the multiselect, first, you should not see the zorga
- Search for the zorga : you should see it. 
Do exactly the same for users. 

## Related issues

- Related to #2449 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [X] Local manual tests


apps/frontend/vitest.config.ts
```diff
@@ -1,13 +1,25 @@
...
 
+const cssStubPlugin: Plugin = {
+  name: 'css-stub',
+  transform(_code, id) {
+    if (id.endsWith('.css')) {
+      return { code: 'export default {}' };
+    }
+  },
+};
+
```
jpkha: Why do we need that ?

apps/frontend/src/components/registration/register/OrganizationForm.test.tsx
```diff
@@ -36,6 +35,17 @@ vi.mock('@filigran/ui/clients', async (importOriginal) => ({
   FormMessage: () => null,
 }));
 
+vi.mock('@filigran/ui/servers', () => ({
```
jpkha: why do we need to mock component ? They shouldn't have any business logic and should be render without issue ?

apps/frontend/vitest.config.ts
hervyt: INdeed, removed it

apps/frontend/src/components/registration/register/OrganizationForm.test.tsx
hervyt: Indeed, I remove the useless mocks in the test files :package:

jpkha (approved):

Issue #2593, opened 2026-06-17
fix(trial): last connectivity check information is displayed even when trial is pending


Kroustille:
## Description

The connectivity check data is displayed on trial details page even when trial is pending. The message is confusing because it's not possible to register trial when platform is not ready.

## Environment

1. OS: { e.g. macOS 14, Windows 11, Ubuntu 22.04, etc. }
2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

Steps to create the smallest reproducible scenario:
1. ask for a trial via XTM Hub form
2. check trial details page

## Expected output

Hide connectivity check message 

## Actual output

`Connection lost. You should register your platform again.` message is displayed when trial is pending

## Additional information

<img width="3024" height="1602" alt="Image" src="https://github.com/user-attachments/assets/349262ed-c548-4e1f-b1f8-a0593678496c" />

PR #2594, opened 2026-06-17
fix(trial): conditionally display connection status and last check (#2593)

Kroustille:
# Context
>

This PR fixes the trial registration details UI so that connectivity information (connection status + last connectivity check) is only shown when a trial’s deployment is **ACTIVE**, preventing confusing “connection lost” messaging while the trial is still pending.

**Changes:**
- Conditionally render the “Connection status” and “Last connection check” rows based on trial deployment status (ACTIVE-only for trials; unchanged for non-trials).
- Adjust the cancel button styling in the status row to better fit the inline layout.
- Add unit tests covering the new conditional display behavior for trials.

- start a trial
- check that connection status is not displayed
- make the trial active
- check that connectivity status is displayed

## Related issues

- Related to #2593 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

## Inactive trial

image

## Active trial

image

### DIFF
--- a/apps/frontend/src/components/service/registration/RegistrationDetails.tsx
+++ b/apps/frontend/src/components/service/registration/RegistrationDetails.tsx
@@ -92,6 +92,11 @@ export const RegistrationDetails = ({
     1
   );
 
+  const isLastConnectivityCheckDisplayed =
+    !isTrial ||
+    platform?.deployment_request?.hub_status ===
+      DeploymentRequestHubStatusEnum.ACTIVE;
+
   return (
     <section className="flex justify-between p-xl border border-solid border-blue rounded">
       <ul className="text-sm flex flex-col gap-l">
@@ -118,6 +123,7 @@ export const RegistrationDetails = ({
             {isCancellable && (
               <Button
                 variant="link-destructive"
+                className="cursor-pointer m-0 p-0 ml-4 h-full"
                 onClick={() => setOpenCancelSheet(true)}>
                 {t('Utils.Cancel')}
               </Button>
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/service/registration/RegistrationDetails.tsx
```diff
@@ -118,6 +123,7 @@ export const RegistrationDetails = ({
             {isCancellable && (
               <Button
                 variant="link-destructive"
+                className="cursor-pointer m-0 p-0 ml-4 h-full"
```
jpkha: why cursor-pointer as a button should not have a cursor pointer. Cursor pointer is for a link. Is it an action or a link ?
By the way the variant we call is` link-destructive` so there may be some issue here.

apps/frontend/src/components/service/registration/RegistrationDetails.tsx
```diff
@@ -173,37 +179,41 @@ export const RegistrationDetails = ({
...
+              {!isConnectionStatusOk && (
+                <span>
+                  {t('Register.Details.ConnectionStatus.NotConnectedDetails')}
+                </span>
+              )}
+            </li>
+            <li>
+              <span className="text-gray/60">
+                {t('Register.Details.LastConnectionCheck')}:
+              </span>{' '}
```
jpkha: Does the space is to make some "spacing" between element ? it would be better to use space-x / gap-x or margin IMO

apps/frontend/src/components/service/registration/RegistrationDetails.tsx
Kroustille: The issue is that it's a button because it's an action to open the cancellation drawer, but it is displayed as a link

jpkha: ok so the correct pattern should cursor-pointer for the link. So the variant link-destructive should apply a cursor-pointer.
-> bug in Filigran-UI side.

A button action displayed as link could be okay but behind is not a link so I would not use a cursor-pointer IMO.

Maybe we can check with the design team, but I'm fine with either approach. I just want to make sure we're aligned on a consistent guideline.

jpkha (approved):

Issue #2567, opened 2026-06-15
fix(perf): n+1 query issue


carinelebas:
## Description

We have performance issues on some requests, mainly:
- publicDocumentsByServiceSlugQuery
- publicDocumentBySlugQuery

These queries have field resolvers, leading to n+1 issues.

## Environment

production

## Reproducible steps
Depending if the pod is colocated with the DB, or of the distance/latency to the DB, the request duration increase significantly.

## Expected output

No more significant increase of the request duration depending on the location.

PR #2615, opened 2026-06-19
fix(perf): batch field resolvers with DataLoader (#2567)

theochhean:
# Context
The `publicDocumentsByServiceSlugQuery` and `publicDocumentBySlugQuery` queries suffer from  from N+1 performance issues. Each document returned triggers individual SQL queries for its field resolvers (`uploader`, `uploader_organization`, `children_documents`, `images`, `use_cases`, `integration_Type`)

This PR implements DataLoader to batch and deduplicate these field resolver queries per request.

From my local testing with 67 documents, I went from 5*67=335 queries, to only 5, and from ~350ms to ~13ms

## Related issues

- Related to #2567

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/backend/src/index.ts
+++ b/apps/backend/src/index.ts
@@ -24,7 +24,7 @@ import { initCronJobs, stopCronJobs } from './crons';
 import { PortalContext } from './model/portal-context';
 import { UserLoadUserBy } from './model/user';
 import { documentDownloadEndpoint } from './modules/document/document-download-endpoint';
-import { createDocumentDataLoaders } from './modules/document/document.dataloader';
+import { DocumentDataLoader } from './modules/document/document.dataloader';
 import { documentVisualizeEndpoint } from './modules/document/visualize-document-endpoint';
 import { initAuthPlatform } from './modules/security-management/authentication/auth-platform';
 import { errorLoggingPlugin } from './server/apollo-plugins/log';
@@ -230,7 +230,6 @@ app.use(function (req, res, next) {
     {
       user: req.session.user,
       correlationId: uuidv4(),
-      dataLoaders: createDocumentDataLoaders(),
     },
     () => {
       next();
@@ -311,8 +310,7 @@ const middlewareExpress = expressMiddleware(server, {
       user: user as UserLoadUserBy,
       req,
       res,
-      dataLoaders:
-        requestContext.get()?.dataLoaders ?? createDocumentDataLoaders(),
+      dataLoaders: DocumentDataLoader.create(),
     };
 
     return portalContext;
--- a/apps/backend/src/index.ts
+++ b/apps/backend/src/index.ts
@@ -24,6 +24,7 @@ import { initCronJobs, stopCronJobs } from './crons';
 import { PortalContext } from './model/portal-context';
 import { UserLoadUserBy } from './model/user';
 import { documentDownloadEndpoint } from './modules/document/document-download-endpoint';
+import { createDocumentDataLoaders } from './modules/document/document.dataloader';
 import { documentVisualizeEndpoint } from './modules/document/visualize-document-endpoint';
 import { initAuthPlatform } from './modules/security-management/authentication/auth-platform';
 import { errorLoggingPlugin } from './server/apollo-plugins/log';
@@ -226,7 +227,11 @@ if (
 
 app.use(function (req, res, next) {
   requestContext.run(
-    { user: req.session.user, correlationId: uuidv4() },
+    {
+      user: req.session.user,
+      correlationId: uuidv4(),
+      dataLoaders: createDocumentDataLoaders(),
+    },
     () => {
       next();
     }
@@ -306,6 +311,8 @@ const middlewareExpress = expressMiddleware(server, {
       user: user as UserLoadUserBy,
       req,
       res,
+      dataLoaders:
+        requestContext.get()?.dataLoaders ?? createDocumentDataLoaders(),
     };
 
     return portalContext;
--- a/apps/backend/src/modules/document/document.dataloader.ts
+++ b/apps/backend/src/modules/document/document.dataloader.ts
@@ -27,23 +27,23 @@ export const DocumentDataLoader = {
   batchLoadUploaders: async (
     ids: readonly string[]
   ): Promise<(User | null)[]> => {
-    const rows = await db<User>('User')
+    type Row = User & { _document_id: string };
+    const rows = await db<Row>('User')
       .leftJoin('Document', 'Document.uploader_id', 'User.id')
       .whereIn('Document.id', ids)
       .select('User.*', 'Document.id as _document_id');
 
-    const map = new Map<string, User>();
-    for (const row of rows) {
-      const docId = (row as User & { _document_id: string })._document_id;
-      map.set(docId, row);
-    }
+    const map = new Map<string, User>(
+      rows.map((row: Row) => [row._document_id, row])
+    );
     return ids.map((id) => map.get(id) ?? null);
   },
 
   batchLoadUploaderOrganizations: async (
     ids: readonly string[]
   ): Promise<(Organization | null)[]> => {
-    const rows = await db<Organization>('Organization')
+    type Row = Organization & { _document_id: string };
+    const rows = await db<Row>('Organization')
       .leftJoin(
         'Document',
         'Document.uploader_organization_id',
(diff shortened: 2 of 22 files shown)


apps/backend/src/index.ts
```diff
@@ -226,7 +227,11 @@ if (
 
 app.use(function (req, res, next) {
   requestContext.run(
-    { user: req.session.user, correlationId: uuidv4() },
+    {
+      user: req.session.user,
+      correlationId: uuidv4(),
+      dataLoaders: createDocumentDataLoaders(),
```
carinelebas: if dataloaders are never accessed outside of resolver layer (where they have access to the portalContext), I don't think we need to add them in the requestContext.

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -0,0 +1,175 @@
...
+import { Document } from './document.helper';
+import { DOCUMENT_IMAGE_METADATA_KEYS } from './document.model';
+import { DocumentMetadataDomain } from './domain/document.metadata.domain';
+
+const batchLoadUploaders = async (
+  ids: readonly string[]
+): Promise<(User | null)[]> => {
+  const rows = await db<User>('User')
+    .leftJoin('Document', 'Document.uploader_id', 'User.id')
+    .whereIn('Document.id', ids as string[])
```
carinelebas: this cast is not necessary (same for other functions)

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -0,0 +1,175 @@
...
+    .select('User.*', 'Document.id as _document_id');
+
+  const map = new Map<string, User>();
+  for (const row of rows) {
```
carinelebas: I think this could be simplified like this, wdyt? (same for other functions)
```typescript
    type Row = User & { _document_id: string };

    const rows = await db<Row>('User')...
      
    const map = new Map<string, User>(
      rows.map((row: Row) => [row._document_id, row])
    );
```

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -0,0 +1,175 @@
...
+    map.set(docId, row);
+  }
+  return ids.map((id) => map.get(id) ?? null);
+};
+
+const batchLoadChildrenDocuments = async (
+  ids: readonly string[]
+): Promise<Document[][]> => {
+  const context = requestContext.get();
+  const query = db<Document>('Document_Children')
```
carinelebas: I'm a bit concerned by the fact that we are mainly duplicating the loadChildrenDocuments function (and same for other functions in fact), and that this one is much less tested than the other. There is a high probability that we will forget to update this one. I think we should rather update the loadChildrenDocument function to support batch loading. 

like this for example. wdyt ? 
``` typescript
const buildChildrenDocumentsQuery = <T extends object = Document>(
   parentIds: string[],
   options: { isDataLoader?: boolean; includeMetadata?: DocumentMetadataKeyCode[] } = {}
 ) => {
   const context = requestContext.get();
   const { isDataLoader = false, includeMetadata = [] } = options;

   const query = db<T>('Document_Children')
     .leftJoin('Document', 'Document.id', 'Document_Children.child_document_id')
     .whereIn('Document_Children.parent_document_id', parentIds)
     .modify((qb) => {
       if (context?.user) restrictDocumentToUserOrganization(qb);
     })
     .orderBy('created_at', 'asc')
     .groupBy('Document.id');

   if (isDataLoader) {
     query
       .select('Document.*', 'Document_Children.parent_document_id as _parent_id')
       .groupBy('Document_Children.parent_document_id');
   } else {
     query.select('Document.*');
   }

   DocumentMetadataDomain.addIncludeMetadataQuery(query, includeMetadata);
   return query;
 };

loadChildrenDocuments: async (documentId, includeMetadata = []) =>
   buildChildrenDocumentsQuery([documentId], { includeMetadata }),

 batchLoadChildrenDocuments: async (ids) => {
   type Row = Document & { _parent_id: string };
   const rows: Row[] = await buildChildrenDocumentsQuery<Row>([...ids], { isDataLoader: true });
   // grouping...
 },
```

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -0,0 +1,175 @@
...
+} from '../../__generated__/resolvers-types';
+import { requestContext } from '../../context/request.context';
+import UseCase from '../../model/kanel/public/UseCase';
+import User from '../../model/kanel/public/User';
+import { restrictDocumentToUserOrganization } from '../../security/restriction/document';
+import { Document } from './document.helper';
+import { DOCUMENT_IMAGE_METADATA_KEYS } from './document.model';
+import { DocumentMetadataDomain } from './domain/document.metadata.domain';
+
+const batchLoadUploaders = async (
```
carinelebas: To be consistent with our other layer formats, I think we should rather have something like:
```
export const DocumentDataLoader = {
batchLoadUploaders: async(...) {..}, 
batchLoadUploaderOrganizations: async(..){...},

create: (): DocumentDataLoaders // the current createDocumentDataLoaders. 
}
```

carinelebas (commented):
great perf improvement! I'm mainly suggesting some refacto to avoid code duplication.

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -120,26 +120,36 @@ export const DocumentDomain = {
...
-      .where('Document.id', '=', documentId)
-      .select('Organization.*')
-      .first();
+      .whereIn('Document.id', documentIds)
+      .select('Organization.*', 'Document.id as _document_id');
+  },
+
+  loadUploaderOrganization: async (
+    documentId: string
+  ): Promise<Organization | undefined> => {
```
carinelebas: [nit] technically the return type is `Promise<WithDocumentId<Organization> | undefined>`

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -120,26 +120,36 @@ export const DocumentDomain = {
...
-      .first();
+      .whereIn('Document.id', documentIds)
+      .select('User.*', 'Document.id as _document_id');
   },
 
-  loadUploaderOrganization: async (
-    documentId: string
-  ): Promise<Organization | undefined> => {
-    return db<Organization>('Organization')
+  loadUploader: async (documentId: string): Promise<User | undefined> => {
```
carinelebas: [nit] technically the return type is `Promise<WithDocumentId<User> | undefined>`

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -158,14 +108,13 @@ export const DocumentDataLoader = {
   batchLoadIntegrationTypes: async (
     ids: readonly string[]
   ): Promise<(IntegrationType | null)[]> => {
-    type Row = { document_id: string; value: string };
-    const rows: Row[] = await db<Row>('Document_Metadata')
-      .select('document_id', 'value')
-      .whereIn('document_id', ids)
-      .where('key', DocumentMetadataKeyCode.IntegrationType);
+    type Row = { document_id: string; value: IntegrationType };
+    const rows: Row[] = await DocumentMetadataDomain.buildIntegrationTypeQuery([
```
carinelebas: nit: you can avoid the spread of ids if `buildIntegrationTypeQuery` has `ids: readonly string[]` as parameter type (same for other calls). It should not impact the other calling functions.

apps/backend/src/index.ts
```diff
@@ -306,6 +310,7 @@ const middlewareExpress = expressMiddleware(server, {
       user: user as UserLoadUserBy,
       req,
       res,
+      dataLoaders: DocumentDataLoader.create(),
```
Kroustille, line 313: Is it necessary to recreate dataLoaders for each new request and use it in requestContext ? Or is it possible to create them only once and call them in resolvers ?

carinelebas (approved):

apps/backend/src/modules/document/document.dataloader.ts
theochhean: did not know that, thanks !

apps/backend/src/index.ts
theochhean: This is the pattern suggested from the [doc](https://github.com/graphql/dataloader#creating-a-new-dataloader-per-request)
We do it because the batch depends on the user

Kroustille (approved):

PR #2620, opened 2026-06-19
feat(homepage): leftMenu (#2571)

carinelebas:
# Context
Make the left menu for the new homepage. 

Note: currently new icons from filigran are dark. It will be fixed before merge.

- Check all links in the menu are correct. 
- Check it smoothly expands/collapses 

## Related issues

- Related to #2571 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

Some code is duplicated with the current private menu. However CSS is not the same, and revamp of the private page will be aligned with the public page, so it seems better to directly make new components. Names may need to be changed when doing private menu.


apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,326 @@
...
+  return (
+    <Link
+      href={href}
+      target={external ? '_blank' : undefined}
+      rel={external ? 'noopener noreferrer' : undefined}
+      className={cn(
+        buttonVariants({
+          variant: 'ghost',
+          className:
+            'h-8 w-full justify-start rounded-none normal-case pl-5 pr-s text-xs font-light',
```
hervyt: **nitpick**: On the main menu you don't have a cursor on hover, but you have one on submenu item. Is it wanted ? 🤔

apps/frontend/app/(public)/[locale]/layout.tsx
```diff
@@ -39,80 +42,90 @@ const RootLayout = async ({
...
-            <Link href={`/login`}>{t('PublicLayout.SignIn')}</Link>
-          </Button>
-        </header>
-        <main className="grow overflow-auto">
-          <div className="container pt-l">{children}</div>
-        </main>
-        <footer className="container text-muted-foreground">
-          <div className="items-center justify-between flex flex-col md:flex-row w-full px-4 py-8 gap-l text-center">
-            <span className="txt-default">
+        <div className="flex flex-row grow min-h-0">
```
Comment (hervyt) [inline apps/frontend/app/(public)/[locale]/layout.tsx]: **nitpick**: Don't need flex-row here

apps/frontend/src/components/menu/PublicMenu.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { useLocalStorage } from 'usehooks-ts';
+
+const PublicMenu = () => {
+  const [open, setOpen] = useLocalStorage<boolean>('public-menu-open', true);
+  const handleOpenMenu = useCallback(() => setOpen((prev) => !prev), [setOpen]);
+
+  return (
+    <aside
+      className={cn(
+        'max-md:hidden z-20 flex h-full flex-col overflow-x-hidden bg-page-background text-muted-foreground duration-300',
```
hervyt, line 19: **issue**: On small screens, we just don't have the menu 😱  No way to go on the wanted places...

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,326 @@
...
+}
+
+const PublicLinkMenu = ({
+  href,
+  icon,
+  text,
+  open,
+  external,
+}: PublicLinkMenuProps) => {
+  const currentPath = usePathname();
```
hervyt, line 58: **nitpick**: we use it a lot of times, can we call the hook on the PublicNavigation and pass it in props to avoid multiple calls ?

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,326 @@
...
+        buttonVariants({
+          variant: 'ghost',
+          className:
+            'h-8 w-full justify-start rounded-none normal-case pl-5 pr-s text-xs font-light',
+        }),
+        currentPath === href && ACTIVE_CLASSES
+      )}>
+      <span className="flex w-4 shrink-0 justify-center">
+        <Icon className="h-4 w-4" />
+      </span>
```
hervyt: **question**: can we reuse MenuItemIcon ?

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,326 @@
...
+};
+
+const PublicNavigation = ({ open }: PublicNavigationProps) => {
+  const t = useTranslations();
+  const locale = useLocale();
+  const isCustomViewsEnabled = useIsFeatureEnabled(
+    FeatureFlagEnum.OPENCTI_CUSTOM_VIEWS
+  );
+  const sections = buildSections(t, locale, isCustomViewsEnabled);
+  const bottomLinks = buildBottomLinks(t, locale);
```
hervyt: **nitpick**: Maybe we should memoize these ?

apps/frontend/src/components/menu/PublicNavigation.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+  it('expanding the OpenCTI accordion shows its sub-links', async () => {
+    const user = userEvent.setup();
+    testRender(<PublicNavigation open={true} />);
+
+    const openctiTrigger = screen.getByRole('button', { name: 'OpenCTI' });
+    await user.click(openctiTrigger);
+
+    await waitFor(() => {
+      expect(openctiTrigger).toHaveAttribute('aria-expanded', 'true');
+    });
```
hervyt: **nitpick**: We could extract these and all similar lines (for OpenAEV, XTM One etc.) in a helper ?

hervyt (commented):
Image

The public icon color + arrows in the Close Menu Item are quite difficult to visualize.... Are you sure about these colors ? 🤔

apps/frontend/src/components/menu/PublicNavigation.tsx
carinelebas: `usePathname()` is a built-in Next.js hook and has no cost being called. Passing it as props in the function doesn't have any real added value, I think only adds noise in the interface. Wdyt?

apps/frontend/src/components/menu/PublicMenu.tsx
carinelebas: We have a specific task for mobile, it will be done later but before we release.

apps/frontend/src/components/menu/PublicNavigation.tsx
carinelebas: There is no need, the inputs are stable primitives and the child components aren't memoized, so useMemo would only add noise.

apps/frontend/src/components/menu/public-navigation.config.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export interface BottomLink {
+  key: string;
+  href: string;
+  icon: ElementType;
+  label: string;
+  external?: boolean;
+}
+
+export const buildSections = (
```
Kroustille: What do you think of transforming this const into a custom hook and directly use `useTranslations` and `useLocale` instead of passing it through props ?

apps/frontend/src/components/menu/public-navigation.config.ts
```diff
@@ -0,0 +1,169 @@
...
+        href: 'https://filigran.io/platform/xtm-one/',
+        label: t('PublicMenu.About'),
+        external: true,
+      },
+      { label: t('PublicMenu.AICatalog'), badge: t('PublicMenu.ComingSoon') },
+    ],
+  },
+];
+
+export const buildBottomLinks = (
```
Kroustille: Same for this const as a custom hook

apps/frontend/src/components/menu/PublicMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { MenuLogo } from '@/components/menu/MenuLogo';
+import PublicNavigation from '@/components/menu/PublicNavigation';
+import { cn } from '@/lib/utils';
+import { LogoFiligranIcon } from '@filigran/icon';
+import { useTranslations } from 'next-intl';
+import { useCallback } from 'react';
+import { useLocalStorage } from 'usehooks-ts';
+
+const PublicMenu = () => {
+  const [open, setOpen] = useLocalStorage<boolean>('public-menu-open', true);
```
Kroustille:
```suggestion
  const [isCollapsed, setIsCollapsed] = useLocalStorage<boolean>('is-public-menu-collapsed', true);
```

apps/frontend/src/components/menu/PublicMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const handleOpenMenu = useCallback(() => setOpen((prev) => !prev), [setOpen]);
```
Kroustille, line 13:
```suggestion
  const handleCollapseMenu = useCallback(() => setIsCollapsed((prev) => !prev), [setIsCollapsed]);
```

apps/frontend/src/components/menu/CollapseMenuButton.tsx
Kroustille, line 8:
```suggestion
isCollapsed: boolean;
handleCollapseMenu: () => void;
```

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,325 @@
...
+    <Icon
+      aria-hidden={true}
+      focusable={false}
+      className="h-4 w-4"
+    />
+  </span>
+);
+
+interface PublicLinkMenuProps {
+  open: boolean;
```
Kroustille, line 44:
```suggestion
  isCollapsed: boolean;
```

apps/frontend/src/components/menu/PublicMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+      )}>
+      <MenuLogo />
+      <div className="flex flex-col flex-1 justify-between min-h-0">
+        <PublicNavigation open={open} />
+        <div>
+          <CollapseMenuButton
+            open={open}
+            handleOpenMenu={handleOpenMenu}
+          />
+          <div
```
Kroustille: The Filigran logo is completely hidden when menu is collapsed, making the collapse button move vertically. I think we should still display the Filigran logo when menu is collapsed as it is done on OpenCTI

https://github.com/user-attachments/assets/c440e170-dae1-46c7-bc16-de4999dbc3c2

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,325 @@
...
+        currentPath === href ? ACTIVE_CLASSES : HOVER_CLASSES
+      )}>
+      <MenuItemIcon icon={icon} />
+      <span className={cn('truncate', open ? 'ml-2' : 'sr-only')}>{text}</span>
+    </Link>
+  );
+};
+
+interface PublicNavigationProps {
+  open: boolean;
```
Kroustille, line 79:
```suggestion
  isCollapsed: boolean;
```

apps/frontend/src/components/menu/PublicNavigation.tsx
```diff
@@ -0,0 +1,325 @@
...
+      {open && (
+        <span className="flex-1 px-s text-left text-foreground text-sm font-light truncate">
+          {section.label}
+        </span>
+      )}
+    </Link>
+  );
+};
+
+const PublicNavigation = ({ open }: PublicNavigationProps) => {
```
Kroustille, line 256: I think we should split components into multiple files instead of having one big file with every components.

What do you think ?

Kroustille (commented):
On OpenCTI, when an accordion is opened, the other opened accordion is closed. We may need the same behaviour here to not have a really long menu

apps/frontend/src/components/menu/PublicMenu.tsx
carinelebas: I can rename the local storage, but I'd rather keep the open logic. Open and isCollapsed are inverted logic, so I think it makes the logic harder to understand after (e.g. !isCollapsed to check if expanded). 
Wdyt?

apps/frontend/src/components/menu/PublicMenu.tsx
carinelebas: same here, I'd rather keep the consistency

apps/frontend/src/components/menu/PublicNavigation.tsx
carinelebas: Having all relating components in the same file is currently the logic on filigran-ui. If all components are used only in this file, I find it easier to manage than having many files. Wdyt?

Kroustille: Yes we can see with the private menu if we need some components from there and split it at this time 😉

Kroustille (approved):

Issue #2340, opened 2026-05-19
Integrate a cookie consent tool


EllynBsc:
**Overall Description**
As we're updating the marketing website cookie policy, we need to integrate cookie consent tool for the Hub

PR #2632, opened 2026-06-22
feat(frontend): add per-service cookie consent management (#2340)

Antoine-Quintin:
# Context
>

[Capture vidéo du 2026-06-22 17-41-50.webm](https://github.com/user-attachments/assets/4fc1e011-b98c-4acc-a93b-de22a8aa3730)

Replaces the legacy `tarteaucitron.js` cookie banner with a custom, per-service
consent solution for XTM Hub (React/Next.js). Third-party scripts are
hard-blocked until the user consents, and consent is stored per service so each
one can be toggled independently (a category is "on" only when all its services
are). Revoking a service purges its cookies.

Includes a banner (Accept all / Reject all / Cookie settings), a preferences
modal (master toggle + per-category toggles + per-service drawer), and a
"Cookie settings" entry in the public footer that reopens the modal.

Also bundles a separate-concern fix in `Copilot.tsx`: the widget no longer
duplicates across mounts / React StrictMode, and the banner offsets it so it no
longer overlaps.

1. Fresh load (clear the `xtmhub_consent` cookie): no GA / HubSpot request fires
   (Network tab), no `_ga` / `__hstc` cookies are set.
2. **Accept all** → GA and HubSpot load; their cookies appear.
3. **Cookie settings → disable Performance and analytics → Save** → `_ga*`,
   `_gid`, `_gcl_au` are purged and GA stops; HubSpot cookies untouched.
4. **Reject all** → nothing loads.
5. Reload after any choice → the banner stays hidden, the choice persists.

Unit tests: `cookie-consent.utils` (normalization, per-category aggregation,
cookie purge incl. wildcard patterns) — all green.

## Related issues

- Related to #2340

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information
- The `Copilot.tsx` fix is bundled but is a different concern/ownership; it's
  isolated in its own commit. Happy to split it into a separate PR if preferred.
- The Copilot duplicate/offset fix is not unit-tested (it depends on an external
  widget's shadow DOM); verified manually. A durable fix would be a CSS offset
  variable exposed by the Copilot team.


apps/frontend/src/components/cookie-consent/CookieConsentPreferences.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [draft, setDraft] = useState<ServiceConsent>(consent);
+  const [openDrawers, setOpenDrawers] = useState<Record<string, boolean>>({});
+  const [wasOpen, setWasOpen] = useState(isPreferencesOpen);
+
+  if (isPreferencesOpen !== wasOpen) {
+    setWasOpen(isPreferencesOpen);
+    if (isPreferencesOpen) {
+      setDraft(consent);
+    }
+  }
```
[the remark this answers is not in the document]
Antoine-Quintin: Intentional: this is the React “adjust state during render” pattern (the guard isPreferencesOpen !== wasOpen + setWasOpen closes the condition on the next render, so no render loop). Moving it back into a useEffect would reintroduce the react-hooks/set-state-in-effect error that CI enforces.

apps/frontend/src/components/cookie-consent/cookie-consent.consts.ts
carinelebas, line 1: I think we should call this file `cookie-consent.consts.ts` to be consistent with the codebase

apps/frontend/src/components/cookie-consent/CookieConsentBanner.tsx
```diff
@@ -0,0 +1,87 @@
+'use client';
+
+import { useConsent } from '@/components/cookie-consent/use-consent';
+import { Button } from '@filigran/ui';
+import { useTranslations } from 'next-intl';
+import { useEffect } from 'react';
+
+const COPILOT_HOST_SELECTOR = 'div#filigran-copilot-widget';
```
carinelebas: is this based on `COPILOT_SCRIPT_ID` ? in this case, it should rather be built from this value rather than hardcoded in case it changes

apps/frontend/src/components/cookie-consent/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+interface CookieConsentContextValue {
+  consent: ServiceConsent;
+  isServiceAllowed: (serviceId: string) => boolean;
+  isCategoryAllowed: (category: ConsentCategory) => boolean;
+  showBanner: boolean;
+  isPreferencesOpen: boolean;
+  openPreferences: () => void;
+  closePreferences: () => void;
+  acceptAll: () => void;
+  rejectAll: () => void;
```
carinelebas, line 40: are these functions needed in the context? should it not be simply in cookieConsentPreferences component?

apps/frontend/src/components/cookie-consent/CookieSettingsLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { useConsent } from '@/components/cookie-consent/use-consent';
+import { useTranslations } from 'next-intl';
+
+export const CookieSettingsLink = () => {
+  const t = useTranslations('CookieConsent');
+  const { openPreferences } = useConsent();
+
+  return (
+    <button
```
carinelebas: you should use Filigran-ui Button component

apps/frontend/src/components/cookie-consent/CookieConsentPreferences.tsx
```diff
@@ -0,0 +1,181 @@
...
+                <div className="flex items-start justify-between gap-4">
+                  <div className="flex flex-col gap-1">
+                    <span className="text-sm font-medium text-foreground">
+                      {t(`Categories.${category}.Title`)}
+                    </span>
+                    <span className="text-sm text-muted-foreground">
+                      {t(`Categories.${category}.Description`)}
+                    </span>
+                    {services.length > 0 ? (
+                      <button
```
carinelebas: you should use filigran UI Button

apps/frontend/src/components/cookie-consent/cookie-consent.server.ts
```diff
@@ -0,0 +1,10 @@
+import { CONSENT_COOKIE_NAME } from '@/components/cookie-consent/cookie-consent.registry';
+import { type StoredConsent } from '@/components/cookie-consent/cookie-consent.types';
+import { parseStoredConsent } from '@/components/cookie-consent/cookie-consent.utils';
+import { cookies } from 'next/headers';
+import 'server-only';
+
+export const readServerConsent = async (): Promise<StoredConsent | null> => {
+  const cookieStore = await cookies();
+  return parseStoredConsent(cookieStore.get(CONSENT_COOKIE_NAME)?.value);
```
carinelebas, line 9: The cookie is written client-side with encodeURIComponent, and readClientConsent correctly decodes it — but readServerConsent doesn't. Next.js's cookies() API returns the raw(encoded) value, so JSON.parse throws, returns null, and every returning user sees the consent banner on every page load.

apps/frontend/src/components/external/Copilot.tsx
```diff
@@ -1,12 +1,13 @@
...
 import { usePathname } from 'next/navigation';
-import { useCallback, useEffect, useRef } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { buildContext, CopilotUser, getUserKey } from './copilot.utils';
 
 const COPILOT_WIDGET_URL =
   'https://copilot.filigran.ai/api/v1/public/widget.js';
 const COPILOT_TOKEN = 'jNJu1JTbbPwNqk1tqEOw-WjsKU4dEcgn';
 const COPILOT_SCRIPT_ID = 'filigran-copilot-widget';
+let widgetOwner: symbol | null = null;
```
carinelebas, line 12: I got this claude comment, I'm not sure what to think about it 🤔 
This module-global is fragile: React 18 Strict Mode double-mounts components, and concurrent renders could silently no-op. A cleaner approach is checkingdocument.getElementById(COPILOT_SCRIPT_ID) directly in a useEffect, or storing the dedup flag in a top-level React ref.

apps/frontend/src/components/cookie-consent/use-consent.ts
carinelebas, line 1: Could this file be an export in CookieConsentProvider rather than a separate file?

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  type ConsentRegistry,
+  type ServiceConsent,
+  type ServiceDefinition,
+  type StoredConsent,
+} from '@/components/cookie-consent/cookie-consent.types';
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+
+export const getAllServices = (
+  registry: ConsentRegistry = CONSENT_REGISTRY
```
carinelebas: is it really needed to pass this registry in parameter to every function ? why don't we simply use CONSENT_REGISTRY directly?

apps/frontend/src/components/cookie-consent/CookieConsentProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export const CookieConsentProvider = ({
+  initialConsent,
+  children,
+}: CookieConsentProviderProps) => {
+  const [stored, setStored] = useState<StoredConsent | null>(initialConsent);
+  const [isPreferencesOpen, setIsPreferencesOpen] = useState(false);
+
+  const consent = useMemo(
+    () => (stored ? normalizeConsent(stored.services) : getDefaultConsent()),
```
carinelebas: if content is stored normalized, I don't think there is any need to normalize it here?

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
```diff
@@ -0,0 +1,161 @@
...
+      typeof parsed.timestamp !== 'string' ||
+      typeof parsed.services !== 'object' ||
+      parsed.services === null
+    ) {
+      return null;
+    }
+    return {
+      version: parsed.version,
+      timestamp: parsed.timestamp,
+      services: normalizeConsent(parsed.services),
```
carinelebas, line 107: if content is stored normalized, I don't think there is any need to normalize it here?

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export const serializeConsent = (
+  services: ServiceConsent,
+  version: number = CONSENT_VERSION,
+  now: Date = new Date()
+): string =>
+  JSON.stringify({
+    version,
+    timestamp: now.toISOString(),
+    services: normalizeConsent(services),
```
carinelebas, line 122: services has already been normalized before, no ?

apps/frontend/src/components/cookie-consent/cookie-consent.types.ts
```diff
@@ -0,0 +1,31 @@
+export const CONSENT_CATEGORIES = [
```
carinelebas: this is a const, not a type, so it should rather be in the renamed registry file

apps/frontend/src/components/cookie-consent/cookie-consent.server.ts
Antoine-Quintin: Good catch — I checked the runtime behavior: after "Accept all" + a hard reload, the banner does not reappear, so readServerConsent is reading the cookie correctly. Next.js's cookies().get().value returns the decoded value (Next handles percent-decoding), so parseStoredConsent gets valid JSON and JSON.parse succeeds. Adding decodeURIComponent here would actually double-decode and break it. The e2e fixture is consistent with this (writes the encoded value, server decodes on read).

apps/frontend/src/components/cookie-consent/CookieConsentProvider.tsx
Antoine-Quintin: These live in the context because they're shared across several consumers, not just the Preferences component. acceptAll / rejectAll are called from the banner (the "Accept all" / "Reject all" buttons), openPreferences is used by both the banner and the footer "Cookie settings" link, and closePreferences / save by the dialog. Putting them in the Preferences component would make them unreachable from the banner and the footer link. The context is the single owner of consent state, so all mutations go through it — that's also what keeps the cookie write + script gating in one place.

apps/frontend/src/components/external/Copilot.tsx
Antoine-Quintin: Fair point, the module-global was added to handle the widget being mounted from two places (the public layout and MeContext) — without a shared owner, both mounts race and you get duplicate widgets. The widgetOwner symbol dedups across those mounts; in Strict Mode the double-invoke is dev-only and the cleanup handles it. That said, I agree it's the most fragile part here and it's really a Copilot concern, not cookie-consent — the only reason it's in this PR is that the consent banner needs to offset the widget. I'd prefer to keep the current behavior for now (it does fix the real duplicate-widget bug we saw) and move the Copilot dedup/lifecycle into a dedicated follow-up PR where we can do it properly (e.g. getElementById dedup in a single effect, or lifting it to a provider). Does that work for you?

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
Antoine-Quintin: You're right — the registry param was a default-arg meant for test injection, but no test actually passes a custom registry, so it was unused flexibility (YAGNI). Dropped it across all the functions and they now reference CONSENT_REGISTRY directly.

apps/frontend/src/components/cookie-consent/use-consent.ts
Antoine-Quintin: Yes our other contexts co-locate their hook (e.g. useServiceContext in ServiceContext.tsx), so I've moved useConsent into CookieConsentProvider.tsx and removed the separate file to match.

apps/frontend/src/components/cookie-consent/CookieConsentProvider.tsx
Antoine-Quintin: Right — stored already comes from parseStoredConsent, which normalizes, so this was redundant. Simplified to stored ? stored.services : getDefaultConsent().

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
Antoine-Quintin: True that the current callers pass an already-normalized value, but I kept it for symmetry: parseStoredConsent normalizes on read, serializeConsent on write, so whatever gets persisted is always clean even if a future caller passes a partial object.

apps/frontend/src/components/cookie-consent/cookie-consent.utils.ts
Antoine-Quintin: This one I'd keep on purpose: parseStoredConsent is the single entry point for any stored value (server read, client read, post-write), so the cookie can be stale or hand-edited — an older version, or a service no longer in the registry. Normalizing here guarantees a clean ServiceConsent whatever is in the cookie. The redundant downstream call (the provider one) is gone, but this is the defensive boundary on read.

carinelebas (approved):

Issue #2627, opened 2026-06-22
fix(scope): Display the list of use cases in filter that belongs to the OCTI library (not others)


EllynBsc:
## Description

**Given output/problem:**

- In OpenCTI integrations, it seems that we have the use cases of a lot of libraries instead just the 11 use cases we defined.
- Same for OpenCTI custom dashboards.
- OpenCTI playbooks has 15 use cases and not 11

**Expected output:** 
The use cases of the libraries need to display the use cases that belongs to the library

1. Adversary & Campaign Insights
2. Vulnerability & Exploit Awareness
3. Infrastructure & Attack Surface Visibility
4. Detection & Response Enablement
5. Fraud, Financial Crime & Cryptocurrency Monitoring
6. Brand, Digital Risk & Underground Exposure
7. Third-Party & Supply Chain Oversight
8. Cloud, SaaS & Platform Security
9. Geopolitical, Physical & Hybrid Risk Analysis
10. Vertical Market & Mission-Specific Intelligence
11. Foreign Influence and Disinformation (FIMI)
12. Data & Health Administration

See notion here: https://app.notion.com/p/filigran/Use-cases-for-XTM-Hub-Integrations-2e08fce17f2a806bbe99c6d5fb6d76ad

PR #2770, opened 2026-07-07
feat(use-case): add filigran product in UseCase table (#2627)

jpkha:
# Context
>

Add new column of Product to be able to filtered out the use case depending of the Product services

1. Go to Use Case in admin panel
2. Create use Case and specify a product like OpenCTI
3. Go to a OpenCTI like dashboard, integrations
4. Try to create or modify, you should be able choose UseCase that is belong to OpenCTI

## Related issues

- Related to #2627 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/setup-vitest.ts
```diff
@@ -78,6 +78,22 @@ setDefaultGlobalMocks();
...
+
+Object.defineProperty(globalThis, 'ResizeObserver', {
+  writable: true,
+  value: ResizeObserverMock,
+});
+
+Object.defineProperty(window.HTMLElement.prototype, 'scrollIntoView', {
+  writable: true,
+  value: vi.fn(),
+});
```
jpkha, line 95: Initializing the Multiselect component is causing some issues, so we need to mock some events to avoid these errors when rendering it.

apps/backend/src/modules/common/common.graphql
```diff
@@ -34,4 +41,12 @@ input LogicalFilterInput {
 
 type Success {
   success: Boolean!
+  warnings: [String!]
+  invalidUseCasesByConnector: [InvalidUseCasesByConnector!]
```
Kroustille: This property seems to be a specific property inside a generic type and does not seem to be used or returned within code 🤔

Kroustille (approved):

apps/backend/src/modules/common/common.graphql
jpkha: You are right. They are a mixing in the code with my other PR. Will clean up before merging thank you !

PR #2771, opened 2026-07-07
feat(homepage): display connected products section on homepage (#2665)

Kroustille:
# Context
>

This PR enhances the authenticated homepage by introducing a “connected products” section that surfaces registered platforms (and trial requests) as cards, adds a cross-sell CTA to try the other product when applicable, and personalizes the welcome label using the current user’s name.

**Changes:**
- Add homepage view-model utilities to map `registeredPlatforms` into UI cards and compute cross-sell eligibility.
- Introduce new homepage UI components (registered platform cards + “try other product” block) and render them from `PrivateHomepage`.
- Extend GraphQL queries/types and i18n messages to support the new homepage content.

- check display when there is 0 registered product
- request a trial
- check the trial is displayed and the CTA is linked to the other product
- register a platform with the other product
- check that the registered platform is displayed

## Related issues

- Related to #2665

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/src/components/homepage/Homepage.utils.ts
```diff
@@ -1,19 +1,51 @@
...
   ServiceDefinitionIdentifier,
 } from '@graphql/generated';
 
+type RegisteredPlatformForHomepage =
+  RegisteredPlatformsQuery['registeredPlatforms'][number];
+
+export type HomepageRoadmapTitleProduct = 'opencti' | 'openaev' | 'default';
+
+export type HomepageRoadmapResolution = {
+  productFilter: FiligranProductEnum | undefined;
```
carinelebas: I think it would be better to use FiligranProduct instead of FiligranProductEnum. It would be better to avoid using script generated enums in the new code, wdyt?

apps/backend/src/modules/document/document.resolver.test.ts
```diff
@@ -552,7 +558,7 @@ describe('document GraphQL query', () => {
 
     const result = await documentResolver.Query!.document!(
       {},
-      { documentId: docId },
+      { documentId: docId, serviceInstanceId: SERVICES.INSTANCES.EPIC.ID },
```
carinelebas: Do we really call this endpoint with EPIC serviceInstanceId? I think even in tests we should use valid data.

apps/frontend/src/components/homepage/Homepage.utils.ts
```diff
@@ -1,19 +1,43 @@
...
+  titleProduct: HomepageRoadmapTitleProduct;
+};
+
+export type HomepageRegisteredProduct = 'opencti' | 'openaev';
+
+export type HomepageRegisteredPlatformCardViewModel = {
+  id: string;
+  product: HomepageRegisteredProduct;
```
carinelebas: why don't we reuse platformIdentifier here? it would avoid a lot of other mappings?

apps/frontend/src/components/homepage/RegisteredPlatformCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { PlatformContractEnum } from '@generated/models/PlatformContract.enum';
+import { PlatformContract } from '@graphql/generated';
+import { useFormatter, useTranslations } from 'next-intl';
+import React from 'react';
+
+type RegisteredPlatformCardProps = {
+  platform: HomepageRegisteredPlatformCardViewModel;
+};
+
+const PRODUCT_LABEL_BY_PRODUCT = {
```
carinelebas: if we use the platformIdentifier, this is already defined in PlatformMetadataMapping

apps/frontend/src/components/homepage/RegisteredPlatformCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+  opencti: 'OpenCTI',
+  openaev: 'OpenAEV',
+} as const;
+
+const CONTRACT_LABEL_BY_CONTRACT: Record<PlatformContractEnum, string> = {
```
carinelebas: I wonder if we should not also define more generally these values? it may not be specific to this, and it's already used in Register.Details.Contracts for example...

apps/frontend/src/components/homepage/TryOtherPlatformProductBlock.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const TryOtherPlatformProductBlock = ({
+  product,
+}: TryOtherPlatformProductBlockProps) => {
+  const t = useTranslations('HomePage.TryOtherPlatform');
+
+  const href = PlatformMetadataMapping[product].learnMorePrivateUrl;
+  const productKey = PlatformMetadataMapping[product].name;
+  const ProductIcon =
```
carinelebas: question: would it be worth adding this in `PlatformMetadataMapping`? 
Also since it's used here: https://github.com/FiligranHQ/xtm-hub/pull/2771/changes#diff-03d3bff8a4a800b81afe94d8caafebec8f988cd809b0867fa3480cb8ca823460R82

apps/frontend/src/components/homepage/Homepage.utils.test.ts
```diff
@@ -1,92 +1,191 @@
...
+    vi.setSystemTime(now);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it.each`
+    endDate                       | expected     | description
+    ${null}                       | ${undefined} | ${'returns undefined when end date is missing'}
```
carinelebas: won't `${undefined}` be cast as string? (copilot comment)

apps/frontend/src/components/homepage/Homepage.utils.test.ts
```diff
@@ -1,92 +1,191 @@
...
-      children_documents: [{ id: 'd1', image_type: DocumentImageType.Logo }],
-      service_instance_id: 'inst-1',
-      expected: '/document/images/inst-1/d1',
-      description:
-        'returns the image URL when logo and service_instance_id are both present',
-    },
-  ])(
+  it.each`
+    children_documents                                          | service_instance_id | expected                        | description
+    ${null}                                                     | ${'inst-1'}         | ${undefined}                    | ${'returns undefined when children_documents is null'}
```
carinelebas: why this change? won't undefined and null be treated as string (copilot comment)? 
same question for resolveHomepageCrossSellProduct ?

apps/frontend/graphql/registration/registration.query.graphql
```diff
@@ -2,5 +2,21 @@ query RegisteredPlatforms($input: RegisteredPlatformsInput!) {
   registeredPlatforms(input: $input) {
     id
     identifier
+    title
```
carinelebas, line 5: do we need to retrieve all these data? especially in deployment_request?

apps/frontend/src/components/homepage/PrivateHomepage.tsx
```diff
@@ -1,51 +1,74 @@
...
+];
+
+export const PrivateHomepage = async () => {
   const authenticatedClient = await getAuthenticatedGraphqlClient();
 
-  const registeredPlatformsData = await useRegisteredPlatformsQuery.fetcher(
-    authenticatedClient,
-    {
+  const [registeredPlatformsData, meData] = await Promise.all([
+    useRegisteredPlatformsQuery.fetcher(authenticatedClient, {
```
carinelebas, line 30: Note: when we fetch onlyActive platforms, we will also retrieve trials which registration is active. However, The status is changed to inactive only when the `actual_state` is set to removed (meaning platform has actually been deactivated). If we base the display of the instance only on this field only, the trial may still appear in the list when it has just been cancelled/expired, but not yet removed. 
I don't know if we want to be completely accurate 🤔  ?

carinelebas (approved):

apps/frontend/src/components/homepage/PrivateHomepage.tsx
Kroustille: We can keep it like this for the MVP: there shouldn't be too much time between `actual_state` change and platform removal 🙂

PR #2774, opened 2026-07-07
chore(script): preventing from using enum from script (#2774)

jpkha:


.github/workflows/dockerbuild-ci.yml
```diff
@@ -77,6 +77,32 @@ jobs:
         working-directory: ${{ matrix.working-directory }}
         run: yarn lint
 
+      - name: Check forbidden enum/relay imports on modified frontend files
```
carinelebas: [nit] maybe we should not mention relay here?

carinelebas (approved):

.github/workflows/dockerbuild-ci.yml
jpkha: Why?

Issue #2785, opened 2026-07-08
feat(one-click-deploy): duplicate deployments into Postgres


Julienger0:
https://app.notion.com/p/filigran/TDR-22-One-click-deploy-events-queries-38a8fce17f2a80008988f618b4856f04?source=copy_link

PR #2786, opened 2026-07-08
feat(one-click-deploy): duplicate deployments into Postgres (#2785)

Julienger0:
# Context
>
Duplicates one-click-deploy events from Elasticsearch into a new Postgres OneClickDeployment table,plus a backfill script for historical data

## Related issues

- Related to #2785

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/scripts/import-one-click-deployments.ts
```diff
@@ -0,0 +1,64 @@
+import { database } from '../../knexfile';
```
carinelebas: I think it would be better to make a migration script, that will automatically insert the data from elastic search into postgresql. This script needs to run only once (re-running it would duplicate the data), and with the migration, it's guaranteed, plus it runs automatically at deployment.

apps/backend/src/migrations/20260707134621_create_one_click_deployment.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  await knex.schema.createTable('OneClickDeployment', (table) => {
```
carinelebas, line 6: I don't think we need to store in postgresql all the data sent in telemetry, it's not the same needs, and many fields can be better retrieved with joins on the right tables, rather than duplicating the data.
We could have only:
- id
- resource_id
- platform_id
- tenant_id
- user_id
- deployed_at

Wdyt? 

Note: a unique platform is identified by the pair (platform_id, tenant_id), since we implemented multi-tenancy for openaev. It would make more sense to have an index on both these fields and not only the platform_id.

apps/backend/src/modules/telemetry/telemetry.types.ts
```diff
@@ -30,6 +30,7 @@ export enum TelemetryEventType {
 }
 
 export interface BaseTelemetryEvent {
+  id?: string;
```
carinelebas: I'm not convinced we should keep a link between the event in elastic search and in postgres with this id. It adds complexity, and we don't really need to link these events.

If you add a field in this object, it will be send to ES. 
- if you add a field in elastic, you should also make an elastic migration to add it to the schema.
- Since this id doesn't exist currently in ES, there is no script to add it to the already existing events? 

Also, if we want to store this id:  since pg is our main database, we should rather insert into postgresql table (which is "synchronous", meaning no queue), get the id, and push that id to elastic, rather than having yet another id. 

If we need to identify matching events for debugging reasons, we could benefit from having the same timestamp in telemetry and pg table.

If this id is needed for the import script, I think it would be better replaced with a migration script (see other comment).

Or do you see another reason to add this id that I miss?

carinelebas (changes requested):

apps/backend/src/scripts/import-one-click-deployments.ts
Julienger0: Yes it's definitely a way better solution

apps/backend/src/migrations/20260707134621_create_one_click_deployment.js
Julienger0: agreed

apps/backend/src/modules/telemetry/telemetry.types.ts
Julienger0: No you're right, I didn't see things like that

apps/backend/src/migrations/20260708145912_import_one_click_deployments.js
```diff
@@ -0,0 +1,78 @@
+// One-time import of the historical ONE_CLICK_DEPLOY telemetry events from
```
carinelebas: I wonder how this will run on a brand new (feature) env. Is elastic already up and running? 
For standard deployments, es is not restarted, so it should work. But if elastic is not yet running, this will block the server start? An option could be to make this script an elastic search migration (we are more sure that pg is running)

apps/backend/src/es-migrations/1783581050891-import-one-click-deployments.ts
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+import { database } from '../../knexfile.js';
```
carinelebas: you can't import knexfile in es migration, you should do like in https://github.com/FiligranHQ/xtm-hub/blob/main/apps/backend/src/es-migrations/1758618435926-create_organizations.ts

carinelebas (approved):

PR #2804, opened 2026-07-10
feat(homepage): display last deployed resources section (#2666)

Julienger0:
# Context
>

New lastDeployedOverview GraphQL query returning the latest deployments (document, date, author), scoped to the current organization
Adds organization_id / target_product columns to the OneClickDeployment table (+ index), populated when telemetry events are ingested
Reimport migration that backfills one-click deploys from Elasticsearch with these new fields 

Server component that fetches the overview per product, plus a client component with a product selector and row rendering
Resolves registered products directly from the existing PlatformIdentifier values

https://dev-pr-2804.hub.staging.filigran.io/app
## Related issues

- Related to #2666

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/migrations/20260709133825_add_org_product_to_one_click_deployment.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
```
carinelebas: why do you need these changes ? I don't see these new fields used + I'm sure we need them:
- organization_id can be derived from user_id
- target_product can be linked either to the platform type, or to the type of resource

apps/backend/src/modules/telemetry/one-click-deployment.domain.ts
```diff
@@ -3,8 +3,25 @@ import OneClickDeployment, {
...
+  user_id: string | null;
+  deployedAt: Date;
+}
+
 export const OneClickDeploymentDomain = {
   insert: async (init: OneClickDeploymentInitializer): Promise<void> => {
     await db<OneClickDeployment>('OneClickDeployment').insert(init);
   },
+
+  loadLastDeployed: async (
```
carinelebas: maybe you could make a generic function LoadOneClickDeployments, that takes  filter and  optional orderby and limit parameters? it would be more easily reuseable if needed? 
It's true we don't select only the needed fields with this solution, but the overhead is limited, and it is more easily maintainable and in line with the rest of the code? 
Wdyt?

apps/backend/src/modules/document/document.app.ts
```diff
@@ -666,6 +669,53 @@ export const DocumentApp = {
     );
   },
 
+  loadLastDeployedOverview: async (limit: number, platformId: string) => {
+    const events = await OneClickDeploymentDomain.loadLastDeployed(
```
carinelebas: we should add a check that the platform is registered in the selected organization of the user making the query. The auth directive only checks that the user is authenticated, not that he has permission over the requested data.

apps/backend/src/modules/document/document.app.ts
```diff
@@ -666,6 +669,53 @@ export const DocumentApp = {
...
+    const uniqueUserIds = [
+      ...new Set(events.flatMap((e) => (e.user_id ? [e.user_id] : []))),
+    ] as UserId[];
+
+    const [documents, users] = await Promise.all([
+      DocumentDomain.loadDocumentsWithMetadataByIds(
+        uniqueResourceIds,
+        ALL_METADATA_KEYS
+      ),
+      uniqueUserIds.length > 0
```
carinelebas: maybe it would be easier to use a field resolver with a data loader to load the user information? Like its done for the uploader in document : https://github.com/FiligranHQ/xtm-hub/blob/main/apps/backend/src/modules/document/document.resolver.ts#L178

apps/backend/src/modules/telemetry/one-click-deployment.domain.test.ts
```diff
@@ -26,20 +37,60 @@ describe('oneClickDeploymentDomain', () => {
...
+        platform_id: PLATFORM_A,
+        organization_id: TEST_ORGANIZATIONS.FILIGRAN.ID,
+      });
+      // Same platform, different triggering org: still belongs to the platform.
+      const platformASecondOrgRow = buildRow({
+        platform_id: PLATFORM_A,
+        organization_id: TEST_ORGANIZATIONS.SECOND_ORGANIZATION.ID,
+      });
+      const platformBRow = buildRow({ platform_id: PLATFORM_B });
+      await OneClickDeploymentDomain.insert(platformARow);
```
carinelebas: question: I wonder if you shouldn't use the testHelper to insert these data, since the aim is not to test the insert domain function 🤔

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const serviceSlug = SHAREABLE_RESOURCE_SERVICE_SLUG_MAPPING[resourceType];
+
+  if (!serviceSlug || !document.slug || !document.service_instance_id) {
+    return null;
+  }
+
+  const url = `/app/service/${SHAREABLE_RESOURCE_SERVICE_DEFINITION_IDENTIFIER_MAPPING[resourceType]}/${document.service_instance_id}/${document.id}`;
+
+  const deployedByName = resolveDeployedByName(resource.deployedBy);
+  const deployedAt = format.dateTime(new Date(resource.deployedAt), {
```
carinelebas: there is a formatDate function in src/utils, maybe it would be better to use this one, and if needed define a new data format to use?

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <ResourceTypeIcon
+            resourceType={resourceType}
+            className="size-6"
+          />
+        </div>
+        <div className="flex items-center gap-s">
+          <span className="content-body-base font-bold truncate">
+            {document.name}
+          </span>
+          {firstUseCase && (
```
carinelebas: you can use the BadgeOverflowCounter component to display the badges and +1:  https://github.com/FiligranHQ/xtm-hub/blob/main/apps/frontend/src/components/homepage/HomepageResourceCard.tsx#L92

apps/frontend/src/components/homepage/LastDeployedResourcesSection.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  if (platforms.length === 0) {
+    return null;
+  }
+
+  const client = await getAuthenticatedGraphqlClient();
+
+  const platformsWithOverview = await Promise.all(
+    platforms.map(async (platform) => {
+      const data = await useLastDeployedOverviewQueryQuery.fetcher(client, {
```
carinelebas: so if we have n platform registered, we will make n graphql calls? why not make a graphql endpoint that supports a list of platforms? but maybe the best would be to use lazy loading, and make the call only when the row is selected and we need to display the information?

apps/frontend/src/utils/shareable-resources/resource-type-icon.tsx
carinelebas, line 1: 💯

apps/backend/src/migrations/20260709133825_add_org_product_to_one_click_deployment.js
Julienger0: right it was a leftover of something I tried but I didn't delete it

apps/backend/src/modules/telemetry/one-click-deployment.domain.ts
Julienger0: done

apps/backend/src/modules/telemetry/one-click-deployment.domain.test.ts
Julienger0: right !

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
Julienger0: done

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
Julienger0: done

apps/backend/src/modules/document/document.app.test.ts
```diff
@@ -1387,4 +1389,33 @@ describe('loadMostDeployedDocuments', () => {
 
     expect(result.map((d) => d.id)).toEqual([id1, id3]);
   });
+
+  describe('loadLastDeployedOverview', () => {
```
carinelebas, line 1392: we should also have a happy path test, no?

apps/backend/src/modules/document/document.app.ts
```diff
@@ -670,18 +672,33 @@ export const DocumentApp = {
   },
 
   loadLastDeployedOverview: async (limit: number, platformId: string) => {
-    const events = await OneClickDeploymentDomain.loadLastDeployed(
-      limit,
-      platformId
+    const registeredPlatforms =
```
carinelebas: Here we are fetching all platforms from db (whatever the organization,  then filtering in the backend. 
Also, `isPlatformInOrganization` is not at all checking if the platform is in the organization since we are retrieving all platforms whatever the organization. 
We should rather filter in the DB directly the platform we want to retrieve, using maybe `loadResolvedConfigurationByPlatform`? (or a new dedicated function if it's doing too many queries for our needs), if we use platformId + tenantId, or loadRegisteredPlatform if using the serviceInstanceId, which already retrieves only the platforms from the user organization.

apps/frontend/src/components/homepage/LastDeployedResourcesSection.tsx
Julienger0: done

apps/backend/src/modules/document/document.app.ts
```diff
@@ -670,18 +672,33 @@ export const DocumentApp = {
   },
 
   loadLastDeployedOverview: async (limit: number, platformId: string) => {
```
carinelebas: OpenAEV being multi tenant, the platform id doesn't identify uniquely the registered platform. 
It should be either the serviceInstanceId, or platformId + tenantId

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
```diff
@@ -77,15 +70,11 @@ const LastDeployedResourceRow = ({
...
-          {firstUseCase && (
-            <Badge className={cn('shrink-0', BADGE_CLASS)}>
-              {firstUseCase.name}
-            </Badge>
-          )}
-          {extraUseCasesCount > 0 && (
-            <Badge className={cn('shrink-0', BADGE_CLASS)}>
-              +{extraUseCasesCount}
-            </Badge>
+          {useCases.length > 0 && (
```
carinelebas: checking the length is not needed, the BadgeOverflowCounter manages an empty list

apps/backend/src/modules/document/document.app.ts
Julienger0: I don't really see the need here: we load that user info from a single place, and it's already a single batched query (WHERE id IN (...))
Am I missing something?

apps/backend/src/utils/error/error.code.ts
```diff
@@ -20,6 +20,7 @@ export enum ForbiddenErrorCode {
   NotAllowedByDeploymentStatus = 'NOT_ALLOWED_BY_DEPLOYMENT_STATUS',
   IntegrationTypeNotManageable = 'INTEGRATION_TYPE_NOT_MANAGEABLE',
   ServiceNotManageable = 'SERVICE_NOT_MANAGEABLE',
+  PlatformNotInOrganization = 'PLATFORM_NOT_IN_ORGANIZATION',
```
carinelebas: Security wise, I don't this we should return such detailed message. It indicates that the platform exists, which is too much information. 
We should rather re-use PLATFORM_NOT_REGISTERED.

apps/backend/src/modules/document/document.app.ts
carinelebas: What you do here is make a unique list of the user ids, fetch the users for that list, and embed it in the returned object. This is exactly what a dataloader do. 
Using directly a dataloader would simplify this code, and also avoid doing this query in case the user data doesn't need to be retrieved by the caller. 
The choice is yours :)

apps/backend/src/modules/document/document.app.ts
Julienger0: ✅

apps/backend/src/modules/document/document.app.ts
Julienger0: ✅

apps/frontend/src/components/homepage/LastDeployedResourceRow.tsx
Julienger0: ✅

apps/backend/src/utils/error/error.code.ts
Julienger0: ✅

apps/backend/src/modules/document/document.app.test.ts
Julienger0: ✅

apps/backend/src/modules/document/document.app.ts
Julienger0: ✅

apps/backend/src/modules/registration/registration.domain.ts
```diff
@@ -275,6 +277,19 @@ export const RegistrationDomain = {
             );
           });
         }
+      })
+      .where(function () {
```
carinelebas, line 281: It would be nice to add tests to cover this new parameter

apps/backend/src/modules/registration/registration.domain.ts
```diff
@@ -275,6 +277,19 @@ export const RegistrationDomain = {
...
+        if (hasDeployedResources) {
+          this.whereExists(function () {
```
carinelebas, line 283: I think we should add a migration to update the OneClickDeployment columns platform_id and tenant_id to uuid instead of text, it would avoid the cast. This the table is new, I think it's better to directly update it to the correct type rather than start using cast in different places.
Also, without the cast, the query could be written in pure knex.

carinelebas (approved):

Issue #2828, opened 2026-07-15
fix(connection): OrganizationId is badly decoded on registration


hervyt:
## Description

Sometimes, when a user tries to connect its product to the XTM Hub, the OrganizationId sent is already decoded and the XTMHub's backend then crash. 

## Environment

1. Present on production. Only 1 user had the problem for now 

## Expected output

If the user sends a decoded organizationId or an encoded one, the registration should work.

PR #2829, opened 2026-07-15
fix(registration): Accept directly organizationId if not encoded (#2828)

hervyt:
# Context
> One user had the problem on production and I had the problem on the local dev, while registration, the organizationId can be sent already decoded. Then, if we decode > We have an error. 
> This PR takes the param organizationId if the decoded organizationId is null. 

## How to test
- You should see no regression on registration. 

## Related issues

- Related to #2828

# Checklist

## Quality

- [X] I added the `skip-feature-env` label if a feature environment is not needed for this PR

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [X] Local manual tests
- [X] Hope

### DIFF
--- a/apps/backend/src/modules/registration/registration.resolver.ts
+++ b/apps/backend/src/modules/registration/registration.resolver.ts
@@ -94,10 +94,23 @@ const resolvers: Resolvers = {
   Mutation: {
     registerPlatform: async (_, { input }) => {
       try {
+        const { id: decodedId, type: decodedType } = fromGlobalId(
+          input.organizationId
+        );
+
+        const ALLOWED_TYPES = [
+          'Organization',
+          'IsPlatformRegisteredOrganization',
+        ];
+
+        const organizationId = ALLOWED_TYPES.includes(decodedType)
+          ? decodedId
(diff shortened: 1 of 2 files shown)


hervyt:
Video while I'm having the bug locally :) 
https://github.com/user-attachments/assets/2964e30a-2eec-4ecb-aa3a-31a14029c5b4

carinelebas (approved):

apps/backend/src/modules/registration/registration.resolver.ts
```diff
@@ -94,10 +94,23 @@ const resolvers: Resolvers = {
...
+
+        const ALLOWED_TYPES = [
+          'Organization',
+          'IsPlatformRegisteredOrganization',
+        ];
+
+        const organizationId = ALLOWED_TYPES.includes(decodedType)
+          ? decodedId
+          : input.organizationId;
+
```
Kroustille, line 109: Wouldn't it be better to use Organization type in `isPlatformRegistered` query so the organization id is rightly encoded ? 🤔

PR #2872, opened 2026-07-21
feat(manifests): implement manifest fetching and download endpoints (#2864)

Antoine-Quintin:
# Context
>

This PR implements the **fetching and download** layer of the manifest distribution
feature (#2849): the HTTP endpoints that expose the aggregated connector manifests
already generated and stored by #2705.

Three public endpoints under `/{product}/{version}/{integrationType}/manifests`:
- `GET .../manifests?count=N` — list manifest metadata (`created_at`, `name`), sorted
  by `created_at DESC`, `count` capped server-side.
- `GET .../manifests/latest` — download the latest manifest (streamed from MinIO).
- `GET .../manifests/{name}` — download a specific manifest by name (streamed).

The endpoints are public and identical for registered and unregistered instances, per
the delivery contract. Files are streamed as-is from MinIO — the Hub aggregates, it does
not modify the payload.

This PR includes input validation, baseline error responses (400/404) and the
path-traversal guard on `{name}`. The 429 response body, the 404-vs-503 storage
distinction and the error-contract consistency pass are intentionally scoped to the
next chunk.

Bruno requests are included under `bruno/manifest/` for reviewers to replay against the
feature environment.

## Related issues

- Related to #2864

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,226 @@
...
+      res.status(404).json({ code: 404, message: 'Manifest not found' });
+      return;
+    }
+
+    const key = ManifestHelper.buildManifestObjectKey(
+      params.product,
+      params.version,
+      latest.name
+    );
+    const body = await MinIOClient.downloadFile(key);
```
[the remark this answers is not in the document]
Antoine-Quintin: Applied the missing validation, factored into a shared helper so /latest and /{name} both validate before building the key. Using 404 (not 500) — a malformed stored name means "no servable manifest", not a server fault.

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,226 @@
...
+  max: MANIFEST_RATE_MAX,
+  standardHeaders: true,
+  legacyHeaders: false,
+});
+
+export const listManifests = async (
+  req: Request,
+  res: Response
+): Promise<void> => {
+  try {
```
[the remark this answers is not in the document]
Antoine-Quintin: Added success + 400 coverage for listManifests

bruno/manifest/ListManifests.bru
```diff
@@ -0,0 +1,20 @@
+meta {
+  name: List manifests
+  type: http
+  seq: 2
+}
+
+get {
+  url: {{baseUrl}}/opencti/7.260604.0/connector/manifests?count=10
```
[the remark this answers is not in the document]
Antoine-Quintin: done

apps/backend/src/modules/shareable-resource/manifest/manifest.domain.ts
```diff
@@ -160,4 +167,49 @@ export const ManifestDomain = {
...
+      .select('created_at', 'name')
+      .where({ product, version, type })
+      .orderBy([
+        { column: 'created_at', order: 'desc' },
+        { column: 'id', order: 'desc' },
+      ])
+      .limit(safeCount);
+  },
+
+  getLatestManifest: async (
```
carinelebas: This function is exactly loadManifests, with limit 1. I don't think it's needed, it will just add more maintenance.

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+const MANIFEST_RATE_MAX = 300;
+
+const manifestRateLimiter = rateLimit({
+  windowMs: MANIFEST_RATE_WINDOW_MS,
+  max: MANIFEST_RATE_MAX,
+  standardHeaders: true,
+  legacyHeaders: false,
+});
+
+export const listManifests = async (
```
carinelebas: I know it's not the case for other "endpoint" files, but we in all our layers, we are exporting the functions through an object. Here we are exporting functions individually. I think we should homogenize and also export the functions through a ManifestEndpoint object, like:
```
export const ManifestEndpoint = {
listManifest : () => {...},
}
```

Or do you see a reason we should not do it?

apps/backend/src/server/endpoints/manifest-endpoint.utils.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  PlatformIdentifier,
+  ManifestType,
+} from '../../__generated__/resolvers-types';
+import { MANIFEST_LIST_DEFAULT_COUNT } from '../../modules/shareable-resource/manifest/manifest.consts';
+
+const MANIFEST_NAME_PATTERN =
+  /^connector-manifest-\d+\.\d+(\.\d+)?(-LTS\.\d+)?-\d{12}$/;
```
carinelebas: This forces LTS to be un upper case. It is really wanted? think we support lts in lowercase in other places in the code?

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+import {MinIOClient} from "../../thirdparty/minio/client";
+import {
+  isIntegrationType,
+  isProduct,
+  isValidManifestName,
+  parseCount,
+} from './manifest-endpoint.utils';
+
+const MANIFEST_RATE_WINDOW_MS = 60 * 1000;
+const MANIFEST_RATE_MAX = 300;
```
carinelebas, line 17: how did you choose this rate limit value? Do you know if we have a log somewhere where we can monitor if we have reached this limit ?

apps/backend/src/server/endpoints/manifest-endpoint.utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+  ManifestType,
+  PlatformIdentifier,
+} from '../../__generated__/resolvers-types';
+import { MANIFEST_LIST_DEFAULT_COUNT } from '../../modules/shareable-resource/manifest/manifest.consts';
+
+describe('isValidManifestName', () => {
+  it.each`
+    name                                             | expected | description
+    ${'connector-manifest-7.260604.0-260526113805'}  | ${true}  | ${'nom standard valide'}
```
carinelebas: descriptions should be in english

apps/backend/src/server/endpoints/manifest-endpoint.utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(isProduct(value)).toBe(expected);
+  });
+});
+
+describe('isIntegrationType', () => {
+  it.each`
+    value                     | expected | description
+    ${ManifestType.Connector} | ${true}  | ${'valeur enum valide (connector)'}
+    ${'injector'}             | ${false} | ${'type hors enum (contrat le liste, pas l\'enum)'}
+    ${'collector'}            | ${false} | ${'type hors enum'}
```
carinelebas: do we really need to test 'injector', 'collector', 'nope'? it seems to me these 3 tests are redundant, no?

apps/backend/src/modules/shareable-resource/manifest/manifest.domain.ts
```diff
@@ -160,4 +167,49 @@ export const ManifestDomain = {
     }
     return recovered;
   },
+
+  loadManifests: async (
```
carinelebas, line 171: These new functions are not tested

apps/backend/src/server/endpoints/manifest-endpoint.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+const buildRequest = (
+  params: Record<string, string>,
+  query: Record<string, unknown> = {}
+) => ({ params, query }) as unknown as Request;
+
+const VALID = {
+  product: 'opencti',
+  version: '7.260604.0',
+  integrationType: 'connector',
```
carinelebas: maybe we should use the enums for product and integrationType instead of raw strings?

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+      count
+    );
+    res.status(200).json({ manifests });
+  } catch (error) {
+    logApp.error('Error while listing manifests', { error });
+    res.status(500).json({ code: 500, message: 'Internal server error' });
+  }
+};
+
+const validateManifestParams = (
```
carinelebas: question: I wonder if this function should not rather be in the util file, so the endpoint file only contains endpoint, and we can unit test this validation ? 🤔

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+      name,
+    });
+    res.status(404).json({ code: 404, message: 'Manifest not found' });
+    return;
+  }
+
+  const key = ManifestHelper.buildManifestObjectKey(product, version, name);
+  const body = await MinIOClient.downloadFile(key);
+  if (!body) {
+    res.status(404).json({ code: 404, message: 'Manifest file not found' });
```
carinelebas, line 183: maybe we should also log something here?

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+    const params = validateManifestParams(req, res);
+    if (!params) return;
+
+    const latest = await ManifestDomain.getLatestManifest(
+      params.product,
+      params.version,
+      params.integrationType
+    );
+    if (!latest) {
+      res.status(404).json({ code: 404, message: 'Manifest not found' });
```
carinelebas: maybe we should log something here?

apps/backend/src/server/endpoints/manifest-endpoint.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    const manifest = await ManifestDomain.getManifestByName(
+      params.product,
+      params.version,
+      params.integrationType,
+      name
+    );
+    if (!manifest) {
+      res.status(404).json({ code: 404, message: 'Manifest not found' });
```
carinelebas: do we want to log something here?

apps/backend/src/server/endpoints/manifest-endpoint.utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    ${['5']}     | ${undefined}                   | ${'tableau rejeté (param dupliqué)'}
+  `('$description', ({ raw, expected }) => {
+    expect(parseCount(raw)).toBe(expected);
+  });
+});
+
+describe('isProduct', () => {
+  it.each`
+    value                         | expected | description
+    ${PlatformIdentifier.Opencti} | ${true}  | ${'valeur enum valide (opencti)'}
```
carinelebas: do we really need to test PlatformIdentifier.OpenCTI and openaev? and opengrc and none? it looks the same test to me?

bruno/manifest/GetLatestManifest.bru
```diff
@@ -0,0 +1,15 @@
...
+}
+
+get {
+  url: {{baseUrl}}/opencti/7.260604.0/connector/manifests/latest
+  body: none
+  auth: none
+}
+
+docs {
+  Télécharge le dernier manifest du triplet. 404 si aucun.
```
carinelebas: docs should be in english (same for other .bru files)

apps/backend/src/server/endpoints/manifest-endpoint.ts
Antoine-Quintin: You're right I'll aligne with the object-export convention used across the other layers. user-picture-endpoint exports individually but it looks like the exception rather than the rule, I'll follow the majority pattern. Worth a follow-up to homogenize user-picture-endpoint too?

apps/backend/src/modules/shareable-resource/manifest/manifest.domain.ts
Antoine-Quintin: Good point, dropped getLatestManifest in favor of loadManifests

apps/backend/src/server/endpoints/manifest-endpoint.utils.ts
Antoine-Quintin: You're right,  the name embeds the lowercase lts version string; the uppercase .LTS. only exists in version_padded. My pattern wrongly required uppercase and would reject real LTS manifests. Rather than just flip the case, I loosened it to a traversal-focused guard (^connector-manifest-[A-Za-z0-9.-]+-\d{12}$) so it no longer re-encodes the version format

carinelebas: I don't think you should accept a too loose format. I think we should match it to the format validated for all manifest related versions : https://github.com/FiligranHQ/xtm-hub/blob/36e3bb653d3523b100dfe4c228439709c9759486/apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts#L4

apps/backend/src/server/endpoints/manifest-endpoint.utils.test.ts
Antoine-Quintin: Agreed, trimmed the redundant cases.

apps/backend/src/server/endpoints/manifest-endpoint.ts
Antoine-Quintin: I carried it over from user-picture-endpoint, but yes, it isn't sized for this traffic profile.

For now we don't have any logging when the limit is hit — that will be handled in the next chunk, where we can add a custom handler that logs the hit and returns a JSON body matching the contract.

Once we have an idea of the polling volume we should be able to tune the window/max.

apps/backend/src/server/endpoints/manifest-endpoint.utils.ts
```diff
@@ -3,10 +3,17 @@ import {
...
 } from '../../__generated__/resolvers-types';
-import { ManifestFragmentHelper } from '../../modules/shareable-resource/manifest-fragment/manifest-fragment.helper';
+import {
+  ManifestFragmentHelper,
+  manifestVersionRegex,
+} from '../../modules/shareable-resource/manifest-fragment/manifest-fragment.helper';
 import { MANIFEST_LIST_DEFAULT_COUNT } from '../../modules/shareable-resource/manifest/manifest.consts';
 
-const MANIFEST_NAME_PATTERN = /^connector-manifest-([^/]+)-(\d{12})$/;
+const versionPattern = manifestVersionRegex.source.replace(/^\^|\$$/g, '');
```
carinelebas: It's a good idea to reuse the exact pattern. Though wouldn't it be easier to maintain to have a `raw_version_pattern = (\d+)\.(\d{1,6})\.(\d+)(?:-lts\.(\d+))`,
and the to use `manifestVersionRegex = /^${raw_version_pattern}$/i` rather than making this replace? 
Replacing in the regex seem pretty fragile, wdyt?

Antoine-Quintin: Good point ! Extracted the body into a shared MANIFEST_VERSION_PATTERN string constant that both the helper and the name pattern compose from so no more .source manipulation, and manifestVersionRegex can go back to being private.

carinelebas (approved):

Issue #2935, opened 2026-07-24
fix(ui): the button learn more is not visible


Antoine-Quintin:
Free trial banner button "learn more" is not visible

<img width="537" height="57" alt="Image" src="https://github.com/user-attachments/assets/61946cf0-7a42-4c1b-9932-a9a33602fa2c" />

PR #2942, opened 2026-07-24
fix(banner): make free trial learn more button visible on brand surface (#2935)

Antoine-Quintin:
# Context
>

image

## Related issues

- Related to #2935

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/frontend/src/components/service/trial-instances/banner/PublicTryFiligranProductsBanner.tsx
+++ b/apps/frontend/src/components/service/trial-instances/banner/PublicTryFiligranProductsBanner.tsx
@@ -14,6 +14,7 @@ import { useLocale, useTranslations } from 'next-intl';
 import Image from 'next/image';
 import Link from 'next/link';
 import { useState } from 'react';
+import { BANNER_ACTION_CLASSES } from './banner-styles';
 
 export const PublicTryFiligranProductsBanner = () => {
   const t = useTranslations();
@@ -59,9 +60,8 @@ export const PublicTryFiligranProductsBanner = () => {
         <DropdownMenuTrigger asChild>
           <div className="flex flex-row items-center mr-xl">
             <Button
-              className="ml-s mr-s text-inherit border-current hover:bg-current/10 focus-visible:ring-current/70"
+              className={BANNER_ACTION_CLASSES}
               variant="secondary"
-              size="sm"
               aria-expanded={menuOpen}>
               {t('Service.Trials.LearnMore.Link')}
               <div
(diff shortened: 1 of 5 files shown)


apps/frontend/src/components/service/trial-instances/banner/PublicTryFiligranProductsBanner.tsx
```diff
@@ -59,10 +60,9 @@ export const PublicTryFiligranProductsBanner = () => {
         <DropdownMenuTrigger asChild>
           <div className="flex flex-row items-center mr-xl">
             <Button
```
carinelebas: instead of extracting the classes, we should extract the button component, especially since it's exactly the same as in the other banner (cf our guidelines from JPK)

apps/frontend/src/components/service/trial-instances/banner/TryFiligranProductsBanner.tsx
```diff
@@ -110,7 +113,10 @@ export const TryFiligranProductsBanner = () => {
       learnMore: (
         <Link
           href={`${settings!.base_url_front}${PlatformMetadataMapping[PlatformIdentifier.Openaev].learnMorePrivateUrl}`}
-          className="ml-xs mr-s underline font-bold">
+          className={cn(
```
carinelebas: same remark here, we could extract the Link component, even if in this same file. I think in this case it's better to duplicate the list of classes with the button, to avoid extracting the classes in a common variable.

apps/frontend/src/components/service/trial-instances/banner/PublicTryFiligranProductsBanner.tsx
Antoine-Quintin: Yes, I extracted into LearnMoreBannerButton, which now owns the whole dropdown (state, trigger, menu items) and takes a single getHref prop, it  was the only real difference between the two banners. ( banner-styles.ts is remove ).

Plus doing so let me drop the <div> that sat between DropdownMenuTrigger asChild and the <Button>. 

Radix was injecting its trigger props onto that div instead of the button, so keyboard handling and aria-expanded were landing on a non-focusable element. They're on the button now, and the manual aria-expanded is no longer needed.

apps/frontend/src/components/service/trial-instances/banner/TryFiligranProductsBanner.tsx
Antoine-Quintin: Done — LearnMoreBannerLink with its own class list, duplicated from the button

carinelebas (approved):

PR #2963, opened 2026-07-27
feat(manifest): harden error handling and rate limiting on manifest endpoints (#2955)

Antoine-Quintin:
# Context
>

Follow-up of #2872, which shipped the manifest retrieval endpoints with baseline
error handling. This PR completes the error contract announced there:

- **Single response format.** Every error response is now written by one helper,
  `sendManifestError`, which produces the `{code, message}` shape defined in the
  Manifest API contract. It replaces thirteen hand-written literals.
- **429.** The rate limiter now has a custom handler that returns a
  contract-compliant body and logs when the limit is reached — the two points
  raised in review on #2872.
- **404 vs 503.** `MinIOClient.downloadFile` no longer swallows every S3 error.
  A missing object still returns `null` (404); a storage failure is propagated
  and answered with 503.

It also fixes a pre-existing bug found along the way: error responses returned
from the download routes inherited the `Cache-Control: public, max-age=31536000,
immutable` header, so a 404 could be cached for a year. `sendManifestError` now
clears it.

The routes are now served through the frontend proxy, so the feature environment
`https://dev-pr-{number}.hub.staging.filigran.io` works. Locally, point
`{{baseUrl}}` at the frontend (`http://localhost:3002`) rather than the backend
port — that is the real path in production.

Most cases are replayable from the Bruno collection under `bruno/manifest/`.
**Run the rate limit test last**: it saturates the window for a minute and would
make everything else fail with 429.

**1. Nominal, and the contract fields**

Run `GenerateManifest`, then `ListManifests`. The response must contain only
`created_at` and `name`:

```json
{"manifests":[{"created_at":"2026-07-21T09:27:01.593Z","name":"connector-manifest-7.260604.0-260526113805"}]}
```

The internal `__typename` field that used to appear here is gone. Then
`GetLatestManifest` streams the file itself, with `Content-Type:
application/json` and `Cache-Control: no-cache`.

**2. Validation — 400**

`ListManifestsInvalidProduct` → `{"code":400,"message":"Invalid product"}`.
Each parameter has its own message, so the caller knows what to fix. A malformed
manifest name is rejected the same way, before any lookup, which also covers path
traversal.

**3. Not found — 404, two distinct cases**

`GetManifestByNameNotFound` uses a well-formed but unknown name →
`{"code":404,"message":"Manifest not found"}`.

If a row exists but its object is missing from storage, the message differs —
`Manifest file not found` — and the server logs an error, since the generation
job uploads before inserting and this should never happen. Both are 404 for the
client; the messages tell the two apart.

**4. Storage failure — 503** *(local only)*
podman stop minio-container

Then `GetLatestManifestStorageDown`:

HTTP/1.1 503 Service Unavailable
{"code":503,"message":"Manifest storage is temporarily unavailable"}

The manifest row still exists, so this is deliberately not a 404. Note the
absence of any `Cache-Control` header: error responses no longer inherit the
one-year immutable directive set by the download route. Restart with
`podman start`.

**5. Rate limit — 429** *(run last)*

for i in $(seq 1 305); do
curl -s -o /dev/null http://localhost:3002/opencti/7.260604.0/connector/manifests
done
curl -s -D - http://localhost:3002/opencti/7.260604.0/connector/manifests

HTTP/1.1 429 Too Many Requests
ratelimit-policy: 300;w=60
ratelimit-remaining: 0
retry-after: 54

{"code":429,"message":"Too many requests, please try again later"}

Check the backend logs: a single `[RATE-LIMIT] Manifest request rate limited`
line for the whole burst. The log is throttled to one entry per minute and per
key, so saturating the endpoint does not flood the output.

## Related issues

- Related to #2955

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
```diff
@@ -0,0 +1,55 @@
...
+    limit: MANIFEST_RATE_MAX,
+    windowMs: MANIFEST_RATE_WINDOW_MS,
+  });
+  if (rateLimitLogThrottle.size > MAX_LOG_RATE_ENTRIES) {
+    for (const [entryKey, loggedAt] of rateLimitLogThrottle) {
+      if (now - loggedAt >= RATE_LIMIT_LOG_INTERVAL_MS) {
+        rateLimitLogThrottle.delete(entryKey);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Antoine-Quintin: Entries are only created from the handler, i.e. after a key has already
exceeded the limit — so each entry costs 300+ requests, and filling the cap
would take tens of thousands. I'd rather not apply the suggestion: clearing
the whole map resets throttling for every key, so the next rejected request
from each one logs again — which amplifies log volume under exactly the
distributed-abuse scenario the throttle exists to dampen.

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const buildManifestRateLimiterOptions = (): Partial<Options> => ({
```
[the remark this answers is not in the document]
Antoine-Quintin: rateLimit() takes Partial<Options>, and Options has every field required —
so returning Options here doesn't compile without an as Options cast. That
cast is what I deliberately avoided: it would claim the object carries every
option when it sets five of them.

On misspelled names: excess property checking applies to object literals
assigned to Partial<T> as well, so a typo is already a compile error today.

apps/backend/src/thirdparty/minio/storage-error.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class StorageUnavailableError extends Error {
+  override readonly cause?: unknown;
+
+  constructor(message: string, options?: { cause?: unknown }) {
+    super(message);
+    this.name = 'StorageUnavailableError';
+    this.cause = options?.cause;
+  }
+}
```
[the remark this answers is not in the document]
Antoine-Quintin: Good catch, applied. Passing `options` to `super()` keeps the native
non-enumerable `cause`, which also removes the need for the explicit
declaration.

apps/backend/src/modules/shareable-resource/manifest/manifest.consts.ts
```diff
@@ -3,6 +3,12 @@ import {
   PlatformIdentifier,
 } from '../../../__generated__/resolvers-types';
 
+// 300 req/min/IP ≈ 300 platforms polling once a minute behind a single egress IP.
```
carinelebas: I don't think this comment is necessary:
- the variable names are self explanatory
- the part about the legitimate traffic is based on what information? + it could change, so this comment may become obsolete
- rate limiting is made to avoid abuse

if we want to make a comment, it should be to describe how this value was chosen (arbitrary?)

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  }) as unknown as Request;
+
+const buildResponse = () => {
+  const json = vi.fn();
+  const status = vi.fn().mockReturnValue({ json });
+  return { res: { status } as unknown as Response, status, json };
+};
+
+describe('buildManifestRateLimiterOptions', () => {
+  it('builds the limiter with the configured window and limit', () => {
```
carinelebas: is this test and the next one really needed? there is no custom logic test, only that a value is set. I don't think there is a real added value to have this test?

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export type ManifestErrorStatus = 400 | 404 | 429 | 500 | 503;
+
+export const sendManifestError = (
+  res: Response,
+  status: ManifestErrorStatus,
+  message: string
+): void => {
+  res.status(status).json({ code: status, message });
```
carinelebas, line 35: You could map here directly the error message to the status code (with maybe the possibility to override if needed). I think it would simplify the code for the caller. Not found error (for example) should return 404 most of the time, if not always?

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
```diff
@@ -0,0 +1,21 @@
+import type { Response } from 'express';
+
+export enum ManifestErrorMessage {
+  InvalidCount = 'Invalid count',
+  InvalidManifestName = 'Invalid manifest name',
+  ManifestNotFound = 'Manifest not found',
+  ManifestFileNotFound = 'Manifest file not found',
```
carinelebas: Maybe we could have only one error for not found? There is no added value to expose externally that we have some inconsistencies in our backend?

bruno/manifest/GetManifestByNameNotFound.bru
carinelebas, line 1: This file is not needed. 
We are not using bruno for validation/regression testing, but as a help if we need to manually call these endpoints. If we have a bru file for a valid name, we can simply replace with an invalid name. 
Same for ListManifestInvalidProduct.bru

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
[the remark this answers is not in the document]
carinelebas: Without cleaning the whole map, I think you can just remove the oldest element from the map, it would prevent the map from growing, wdyt?
`const oldestKey = rateLimitLogThrottle.keys().next().value;`

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.test.ts
Antoine-Quintin: Yes both only assert that a constant was copied into the options object,
so they can't fail unless someone deletes the field which is unlikely.

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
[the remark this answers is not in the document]
Antoine-Quintin: Applied, it's much better than clearing the whole map

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
Antoine-Quintin: Yes, I merged it into a single not-found message, the client can't act differently in either case, and exposing that a row exists without its object leaks internal state for no benefit.

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
Antoine-Quintin: Applied, callers no longer pass a status, it's derived from the message through
a `Record<ManifestErrorMessage, ManifestErrorStatus>`.

I added a separate `sendManifestValidationError` rather than an override
parameter: the messages from `validateManifestParams` are built per field and
aren't part of the enum, but they're always 400. Two named functions felt
clearer than one signature with an optional status.

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
```diff
@@ -4,15 +4,23 @@ export enum ManifestErrorMessage {
...
   TooManyRequests = 'Too many requests, please try again later',
   StorageUnavailable = 'Manifest storage is temporarily unavailable',
   InternalServerError = 'Internal server error',
 }
 
-export type ManifestErrorStatus = 400 | 404 | 429 | 500 | 503;
+type ManifestErrorStatus = 400 | 404 | 429 | 500 | 503;
 
-export const sendManifestError = (
+const STATUS_BY_MESSAGE: Record<ManifestErrorMessage, ManifestErrorStatus> = {
```
carinelebas: what do you think about defining errors and statuses at the same time to ease maintenability? 
Something like 
```
const MANIFEST_ERRORS = {
  InvalidCount: { message: 'Invalid count', status: 400 },
  ...
} as const satisfies Record<string, { message: string; status: ManifestErrorStatus }>;
```

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
```diff
@@ -24,19 +24,30 @@ const logRateLimitThrottled = (ip: string, path: string): void => {
   ) {
     return;
   }
+
+  rateLimitLogThrottle.delete(key);
```
carinelebas, line 25: why do you need to delete the key here just to set it right after?

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
```diff
@@ -24,19 +24,30 @@ const logRateLimitThrottled = (ip: string, path: string): void => {
...
+
   if (rateLimitLogThrottle.size > MAX_LOG_RATE_ENTRIES) {
     for (const [entryKey, loggedAt] of rateLimitLogThrottle) {
       if (now - loggedAt >= RATE_LIMIT_LOG_INTERVAL_MS) {
         rateLimitLogThrottle.delete(entryKey);
       }
     }
+    const oldestKey = rateLimitLogThrottle.keys().next().value;
+    if (
+      rateLimitLogThrottle.size > MAX_LOG_RATE_ENTRIES &&
```
carinelebas: is it really needed to do this check again? I'm not sure there is a real impact if we don't ?

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
Antoine-Quintin: Map preserves insertion order and `set` doesn't move an existing key, so re-inserting keeps insertion order aligned with last-log order — which is what makes the eviction below drop the least recently logged entry.

apps/backend/src/server/endpoints/manifest-endpoint.errors.ts
Antoine-Quintin: Good idea ! It removes the duplicated key list between the enum and the
status map, and `as const satisfies` keeps the statuses type-checked while
preserving literal types.

apps/backend/src/server/endpoints/manifest-endpoint.rate-limit.ts
Antoine-Quintin: You're right. It only avoided evicting one entry in the case where the age
cleanup had already brought us back under the cap, and the cost of not having it
is at most one extra log line — removed.

The remaining undefined check is just for the type: keys().next().value is
string | undefined.

carinelebas (approved):

Issue #2972, opened 2026-07-28
chore: add caching to public homepage requests


Kroustille:
# Context

With the release of the homepage revamp, we saw a significant increase in calls frequency for `EpicCountPerTimelineQuery` and `MostDeployedDocumentsQuery` queries.

The result of this queries won't change very often so we can take cache them to reduce server load.

PR #2987, opened 2026-07-29
fix(cache): fix cache for server fetched react query requests (#2972)

carinelebas:
# Context
>

Cache wasn't working because createPortalGraphqlClient used graphql-request's default fetch instead of Next's patched one. 
Rather than keep patching that (fragile, library-internal workaround), we created serverGraphqlFetch using native fetch directly for the cacheable reads — which made the original override unnecessary, so it was removed. 
We couldn't reuse fetch-fn.ts's networkFetch since it's built for Relay's request/response types, not plain codegen'd documents.

Test on fenv that the queries for epic count and documents on homepage are well cached. Check there is no regression on data fetching.

## Related issues

- Related to #2972

# Checklist

## Quality

- [ x] I consider this work complete and ready for review

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/src/lib/graphql-fetch.utils.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+
+    it('does not recurse into arrays', () => {
+      const input = {
+        tags: ['a', 'b'],
+        items: [{ password: 'should-not-be-scrubbed' }],
+      };
+      const result = scrubSensitiveVariables(input);
+      expect(result).toEqual(input);
+    });
```
Kroustille: I'm wondering if it's a good thing to not scrub properties present in array of objects 🤔

apps/frontend/src/lib/server-graphql-fetch.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const response = await fetch(apiUri, {
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ query: document, variables }),
+    ...options,
+    cache: activeCacheConfig,
```
Kroustille, line 46: The cache property should be removed from `ServerGraphqlFetchOptions` type because it's overriden by the `activeCacheConfig`

Kroustille (approved):

apps/frontend/src/lib/server-graphql-fetch.ts
carinelebas: but activeCacheConfig is computed based on options.cache, so no I don't think it should be removed.

PR #2998, opened 2026-07-30
feat(integration): Add solution Category and license_type in form (#2894)

hervyt:
# Context
> Thsi feature add the solution_category field and the license_type field for all integration. 

## How to test
- Add integration 
- You should see the 2 new types 
- The 2 new types should be displayed then in the details page
- The 2 new types should not be visible on other shareableResources. 

https://github.com/user-attachments/assets/ca855033-770b-4868-bf97-934d7889a752

## Related issues

- Related to #2894

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [X] Integration tests
- [X] E2E tests
- [X] Local manual tests


apps/frontend/src/components/service/form/solution-category.utils.ts
```diff
@@ -0,0 +1,30 @@
+import { portalGraphqlClient } from '@/lib/graphql-client';
+import { formatName } from '@/utils/format/name';
+import {
+  FiligranProduct,
+  OrderingMode,
+  SolutionCategoryOrdering,
+  useSolutionCategoriesListQuery,
+} from '@graphql/generated';
+
+export const getSolutionCategories = (product?: FiligranProduct) => {
```
carinelebas: question: should this not be a hook, and be named useSolutionCategories?

apps/backend/src/modules/document/document.graphql
```diff
@@ -123,6 +124,7 @@ input CreateDocumentInput {
   active: Boolean!
   use_cases: [UseCaseId!]!
   solution_category: SolutionCategoryId
+  license_type: String
```
carinelebas: question: should we have an enum rather than a string? It can only be commercial / free

apps/backend/src/modules/document/document.app.test.ts
```diff
@@ -396,6 +424,48 @@ describe('documentApp', () => {
       });
     });
 
+    it('should persist license_type input in document metadata on update', async () => {
```
carinelebas: I wonder if we should have a dedicated test for each field? couldn't we have one test 'should persist provided data and validate each field in it?

apps/backend/src/modules/document/document.app.test.ts
```diff
@@ -176,6 +176,34 @@ describe('documentApp', () => {
       });
     });
 
+    it('should persist license_type input in document metadata', async () => {
```
carinelebas: same question as for udpate

carinelebas (approved):

apps/frontend/messages/ja.json
```diff
@@ -1008,6 +1008,8 @@
       "SlugLabel": "スラッグ",
       "SlugPlaceholder": "スラッグ",
       "SourceCodeLabel": "ソースコード",
+      "SolutionCategoriesLabel": "解答カテゴリ",
```
carinelebas: Claude said this: 6. Japanese translation looks wrong:  "SolutionCategoriesLabel": "解答カテゴリ"  — 解答 means "answer/solution to a problem" (exam sense), not "solution" in the product/business sense. Likely should use a katakana borrowing like ソリューションカテゴリ, consistent with how other product terms are localized elsewhere in  ja.json .

apps/frontend/src/components/service/form/solution-category.utils.ts
hervyt: I did the same like JPK did for Use-case (I wanted the same way of doing to not have 10 diffrent ways of doing...). Do you think it worth passing it into a hook ? In this case, I will refactor the whole UseCasesField as well

carinelebas: We already have useConnectedPlatforms, useOrgaFreeTrials.. I think it would be better to use hook names for what is functionnaly a hook? I agree it would be great to have the same for use cases and solution categories, but it may also be done in a follow up PR?

hervyt: Done in this PR :)

carinelebas (approved):

PR #3010, opened 2026-07-31
refactor: improve loadDocuments performances and clean resolvers (#2980)

Kroustille:
# Context
>

This PR refactors backend document loading to improve performance by reducing heavy SQL joins/aggregations, introducing batching via DataLoaders, and adding targeted DB indexes, while also simplifying shareable-resource resolvers to rely on the shared `Document` interface resolvers.

**Changes:**
- Added batched loading for `service_instance` and `subscription` via new `DocumentDataLoaders` and supporting domain methods.
- Simplified `DocumentDomain.loadDocuments()` query shape (grouping by `Document.id` only) and moved related fetching to resolvers/DataLoaders.
- Added performance benchmark test helpers/tests and a migration introducing indexes to support `loadDocuments`/parent filtering.

- navigate through public and private library pages and check that there is no regressions

## Related issues

- Related to #2980 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/src/migrations/20260729064500_optimize_load_documents_indexes.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  await knex.raw(`
+    CREATE INDEX IF NOT EXISTS idx_document_children_child_document_id
+    ON "Document_Children" ("child_document_id")
+  `);
+
+  await knex.raw(`
```
carinelebas, line 11: do we really need these 2 indexes? they are pretty much the same, except for update_at field? 
Maybe we could have only one index on: 
      "service_instance_id",
      "type",
      "active",
      "created_at" DESC
      "updated_at" DESC,
I think it could also be used for creation ?

apps/backend/src/modules/document/document.dataloader.ts
```diff
@@ -126,6 +168,62 @@ export const DocumentDataLoader = {
...
+    const parsedKeys = keys.map(parseSubscriptionByServiceInstanceLoaderKey);
+    const organizationIds = [
+      ...new Set(parsedKeys.map(({ organizationId }) => organizationId)),
+    ];
+    const serviceInstanceIds = [
+      ...new Set(parsedKeys.map(({ serviceInstanceId }) => serviceInstanceId)),
+    ];
+
+    const subscriptions =
+      await SubscriptionDomain.loadSubscriptionsByOrganizationAndServiceInstanceIds(
```
carinelebas, line 221: question: here we are losing the mapping organization - serviceInstanceId. I'm wondering if we are losing something/returning more data than we should ? 🤔

carinelebas (approved):

apps/backend/src/modules/document/document.dataloader.ts
Kroustille: By loosing the mapping we could return more data than expected but since we are always fetching documents from a single service instance, this case won't happen.

An option would be to preserve the mapping for the DB query but it will be really verbose without a real impact.

carinelebas (approved):

PR #3024, opened 2026-08-04
feat(manifest): add license_type and contact to fragment ingestion and generated manifest (#2995)

Antoine-Quintin:
# Context
>

image

Adds `license_type`, `contact` and `solution_categories` to manifest fragment
ingestion and to the generated manifest, and displays the contributor contact on
the connector details page. This covers all three fields of #2995.

- **`contact` is the contributor's email or GitHub profile URL**, left empty for
  Filigran-supported integrations. It is rendered as
  plain text.
- **The manifest is the single source of truth for connectors**.
  None of the three fields is carried over from a previous connector version by
  `getConnectorMetadataFromExisting`, so every ingestion rewrites them.

`solution_categories` are resolved by name against the `SolutionCategory`
referential, scoped to the fragment's platform.

1. `NewProductVersion` → `{ "product": "opencti", "version": "7.260600.0" }`
2. `IngestManifestFragments` with an unused fragment version and both fields:
```json
   "version": "7.260309.4",
   "min_version": "7.260507.0",
   "license_type": "commercial",
   "contact": "https://github.com/some-contributor"
```
3. Check the metadata rows:
```sql
   SELECT m.key, m.value
   FROM "Document_Metadata" m
   JOIN "Document" d ON d.id = m.document_id
   WHERE d.slug = 'misp' AND d.version = '7.260309.4'
     AND m.key IN ('license_type', 'contact');
```
4. `GenerateManifest` on `opencti` / `7.260600.0`, then `GetLatestManifest` on
   `/opencti/7.260600.0/connector/manifests/latest` — the contract carries
   `"license_type": "commercial"` and `"contact": "https://github.com/some-contributor"`.

Frontend — connectors created by fragment ingestion are tagged `decoupling` and
excluded from the frontend by `excludeDecouplingTag`, so the display was tested
on a seeded connector by inserting the metadata by hand:

```sql
INSERT INTO "Document_Metadata" (document_id, key, value)
VALUES ('<a seeded connector id>', 'contact', 'contributor@example.com');
```

Checked on both the authenticated page (`/app/service/opencti_integrations/…`)
and the public one (`/en/cybersecurity-solutions/opencti-integrations/misp-intel`),
with an email, with a GitHub URL, and with the row deleted — in which case
nothing is rendered at all, which is the common case.

## Related issues

- Related to #2995

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/kanel.config.cjs
```diff
@@ -41,6 +41,9 @@ const COLUMN_ENUM_MAP = {
   ServiceDefinition: {
     identifier: 'ServiceDefinitionIdentifier',
   },
+  SolutionCategory: {
+    product: 'FiligranProduct',
+  },
```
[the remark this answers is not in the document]
Antoine-Quintin: Fair point on scope, but this isn't optional cleanup. SolutionCategory was
missing from COLUMN_ENUM_MAP while UseCase and Epic — same column, same enum —
are both there. Because of that, running `yarn generate-pg-to-ts` downgrades
`product` from `FiligranProduct[]` to `string[]`, and the committed
SolutionCategory.ts had been hand-written to compensate (it has no
`// @generated` header, unlike every other file in that folder).

I hit this while regenerating Kanel to get ObjectSolutionCategory.ts. So the
SolutionCategory.ts diff here isn't a side effect — it's the file restored to
what the generator actually produces. Anyone regenerating without this fix
silently loses the typing again.

It's in its own commit (`fix(kanel): map SolutionCategory.product to
FiligranProduct`) so it stays easy to review and bisect. Happy to split it into
a separate PR if you'd rather, but it needs to land before the next person runs
Kanel.

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.domain.ts
```diff
@@ -77,6 +79,8 @@ const createConnectorDocument = async ({
     ConnectorWithMetadata,
     'datasheet_url' | 'blogpost_url' | 'demo_url'
   >;
+  licenseType?: string;
```
hervyt: **nitpick**: As license_type is Commercial OR Free,  can we make an enum ?

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.test.ts
```diff
@@ -207,3 +210,144 @@ describe('assertHomogeneousLtsBatch', () => {
     ).toThrow(BadRequestErrorCode.MixedLtsManifestFragments);
   });
 });
+
+describe('normalizeOptionalText', () => {
+  it.each`
+    value          | expected     | description
+    ${'free'}      | ${'free'}    | ${'a plain value'}
```
hervyt, line 217: **nitpick**: We can expect that the "normalize" also "capitalize" the name! ;)

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.test.ts
```diff
@@ -207,3 +210,144 @@ describe('assertHomogeneousLtsBatch', () => {
...
+      expect(result).toBeUndefined();
+    }
+  );
+});
+
+describe('validateAndNormalizeContact', () => {
+  it.each`
+    length                    | description
+    ${MAX_CONTACT_LENGTH - 1} | ${'just below the limit'}
+    ${MAX_CONTACT_LENGTH}     | ${'exactly at the limit'}
```
hervyt, line 252: **praise**: <3

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
```diff
@@ -126,4 +128,34 @@ export const ManifestFragmentHelper = {
   }): string => {
     return `${title}-${version}-logo.png`;
   },
+
+  validateAndNormalizeLicenseType: (
+    value: string | null | undefined
```
hervyt: licenseType should be only Commercial/Free

apps/frontend/src/components/service/document/connector/ShareableResourceConnectorDetails.tsx
```diff
@@ -147,6 +148,12 @@ export const ShareableResourceConnectorDetails = ({
         label={t('Service.ShareableResources.Details.Shares')}>
         <span>{roundToNearest(connectorDetails.share_number)}</span>
       </ShareableResourceDetailItem>
+      {connectorDetails.contact && (
+        <ShareableResourceDetailItem
+          label={t('Service.ShareableResources.Details.ContributorContact')}>
+          <span>{connectorDetails.contact}</span>
```
hervyt, line 154: **question**: Maybe we should capitalize this contact ? 🤔  Wdyt ?

hervyt (commented):
**issue**: You miss the injection work about solution categories 🔢  
Otherwise: really cool PR ! 🐰

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
```diff
@@ -126,4 +128,34 @@ export const ManifestFragmentHelper = {
...
+      throw new Error(BadRequestErrorCode.ContactTooLong);
+    }
+    return normalized;
+  },
+
+  normalizeOptionalText(value: string | null | undefined): string | undefined {
+    if (value === null || value === undefined || value.trim() === '') {
+      return undefined;
+    }
+    return value.trim();
```
carinelebas, line 143: is this function useful? can't you simply use `value?.trim()` ?

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
```diff
@@ -126,4 +128,34 @@ export const ManifestFragmentHelper = {
...
+    }
+    return normalized;
+  },
+
+  validateAndNormalizeContact: (
+    value: string | null | undefined
+  ): string | undefined => {
+    const normalized = ManifestFragmentHelper.normalizeOptionalText(value);
+    if (normalized !== undefined && normalized.length > MAX_CONTACT_LENGTH) {
+      throw new Error(BadRequestErrorCode.ContactTooLong);
```
carinelebas, line 134: if it throws, I'd rather call it 'assert', wdyt?

carinelebas (approved):

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
Antoine-Quintin: Yes, renamed to parseContact (and parseLicenseType) — "parse, don't validate": it throws on invalid input but returns the normalised value, which assert* wouldn't convey. That also keeps assert* in this file for pure checks like assertHomogeneousLtsBatch

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
Antoine-Quintin: They're not equivalent, on a whitespace-only string value?.trim() returns '' while this returns undefined. The difference shows up in the database — undefined stores NULL in Document_Metadata, '' stores an empty string, so we'd lose the distinction between "empty" and "absent". Both parseLicenseType and parseContact rely on that.

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.domain.ts
Antoine-Quintin: now typed with the existing LicenseType enum, and LICENSE_TYPES is derived from Object.values(LicenseType) so the list can't drift from the schema.

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
Antoine-Quintin: Renamed to parseContact (and parseLicenseType): it throws on invalid input but also returns the normalised value, which assert* wouldn't convey. I kept that prefix for pure checks like assertHomogeneousLtsBatch.

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
Antoine-Quintin: They're not exactly equivalent, on a whitespace-only string value?.trim() returns '' while this returns undefined. The difference reaches the database — undefined stores NULL in Document_Metadata, '' stores an empty string, so we'd lose the distinction between "empty" and "absent". Both parseLicenseType and parseContact rely on it.

apps/backend/src/modules/shareable-resource/manifest-fragment/manifest-fragment.helper.ts
Antoine-Quintin: I kept the parameter string | null | undefined on purpose — it's the raw manifest value, so it can't be typed before it's validated. The narrowing happens on the return: parseLicenseType now returns LicenseType | undefined and throws InvalidLicenseType otherwise, with LICENSE_TYPES derived from Object.values(LicenseType).

apps/frontend/src/components/service/document/connector/ShareableResourceConnectorDetails.tsx
Antoine-Quintin: We'd rather not I think, contact is either an email address or a GitHub profile URL. Email local parts are case-sensitive, and URL paths always are, so capitalising could display a value that no longer resolves.

hervyt (approved):

apps/backend/src/modules/shareable-resource/manifest/manifest.types.ts
```diff
@@ -12,6 +12,9 @@ export interface ManifestContract {
   source_code: string | null;
   manager_supported: boolean;
   support_version: string | null;
+  license_type: string | null;
```
carinelebas: this could also be the enum, wdyt?

apps/backend/src/modules/shareable-resource/opencti/integration/integration.graphql
```diff
@@ -179,6 +179,7 @@ type Connector implements Node & Document & Integration {
   demo_url: String
   minimum_deployable_version: String
   license_type: String
```
carinelebas: I know it's not your code, but I'm wondering if this should not also be an enum?

apps/backend/src/modules/shareable-resource/manifest/manifest.types.ts
Antoine-Quintin: Done, LicenseType | null. Typing it surfaced the rest of the chain: the connector model and createConnectorDocument were still passing string, so those are typed too.

apps/backend/src/modules/shareable-resource/opencti/integration/integration.graphql
Antoine-Quintin: Done, all 8 declarations (the Integration interface and its 7 implementing types). Generated mocks now produce real enum values instead of random strings, which is a nice side effect.

carinelebas (approved):

Issue #1992, opened 2026-03-30
[backend] Remove sandbox apollo from envs


hervyt:
Remove sandbox apollo from environments

PR #3038, opened 2026-08-05
fix: cnpg backups bareman plugin (#1992)

pierremahot:
# Context
Fix cnpg bareman plugin backup configuration

## How to test
tested intensively on dev cluster

## Related issues

- Related to https://github.com/FiligranHQ/platform/issues/1992

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/chart/templates/postgres.yml
+++ b/chart/templates/postgres.yml
@@ -25,8 +25,9 @@ spec:
   {{- end }}
   {{- if .Values.postgres.backup.enabled }}
   plugins:
-    - name: barman-objectstore
-      spec:
-        objectStoreRef:
-          name: "{{ .Release.Name }}-barman-objectstore"
+    - enabled: true
+      isWALArchiver: true
+      name: barman-cloud.cloudnative-pg.io
+      parameters:
+        barmanObjectName: "{{ .Release.Name }}-barman-objectstore"
   {{- end }}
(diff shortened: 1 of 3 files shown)


chart/templates/postgres.yml
```diff
@@ -23,7 +23,11 @@ spec:
   externalClusters:
     {{- .Values.postgres.spec.externalClusters | toYaml | nindent 4 }}
   {{- end }}
-  {{- if .Values.postgres.spec.plugins }}
+  {{- if .Values.postgres.backup.enabled }}
```
efaure: From my understanding, this remove the possibility to add plugins with postgres.spec.plugins. 
In this case, it would be relevant to fully remove it from values https://github.com/XTM-Hub/xtm-hub/blob/main/chart/values.yaml#L80

@carinelebas  (not so random ping) : I think I remenber you had a use case where you  thought about adding plugins to postgres. But I think the idea was dropped.  As it's not used for now, it seems safe to drop the config (except you tell me you plan to use it soon). If needed in the future we can always re add it.

efaure (approved):

chart/templates/postgres.yml
pierremahot: i have added possibility to have additional plugins defined

carinelebas: @efaure I confirm for now we are not planning to add plugins, but good to know that we still can 😄

PR #3043, opened 2026-08-06
feat(services): hide non-public services from public and private menus (#3042)

Julienger0:
# Context
>

a service whose public flag is set to false:
It disappears from the left menu, both for signed-in users and for visitors on the public site.
Its page returns an error for anyone whose organization has not been granted access
previously, opening the link was enough to gain access automatically.
Its resources no longer show up on the home pages ("Newest" and "Most deployed"), on the public solution pages, or in the sitemap.
Organizations you grant access to see it and use it exactly as any other service.
Platform administrators still see it in Settings → Service, so they can manage it but it stays out of their own workspace menu, which keeps reflecting what they actually have access to.

**feat(security): add service instance visibility restrictions** : Introduces the rules that determine which services a person is allowed to see.
**feat(service-instance): filter service instances by accessibility** :Service listings now only return the services the user can actually access.
**fix(document): restrict documents to accessible service instances** : Documents belonging to restricted services are no longer returned.
f**ix(seo): exclude non-public services from sitemap and public pages** : Non-public services no longer show up in the sitemap or on public pages.
**feat(navigation): hide inaccessible services from menus** : Menus and navigation only display the services available to the user.

go to https://dev-pr-3043.hub.staging.filigran.io/app ,for this PR I hid custom views library so you shouldn't see it n the menu
in https://dev-pr-3043.hub.staging.filigran.io/app/admin/service/U2VydmljZUluc3RhbmNlOjc2ZTg4YjcxLTQwZTItNDY0OS04NTc3LTRmNzI3NTY3M2Q0NQ== 
you can add your organization
once added the custom view library should appear on the left menu (make sure you're in the right workspace)
## Related issues

- Related to #3042

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/document/document.app.ts
```diff
@@ -773,6 +748,27 @@ export const DocumentApp = {
   },
 };
 
+const resolveVisibleDocumentTypes = async (
```
Kroustille: There are missing tests associated with the changes of this file

apps/backend/src/modules/service/definition/service-definition.domain.test.ts
```diff
@@ -1,12 +1,51 @@
...
+
+    beforeEach(() => {
+      originalEnvironment = portalConfig.environment;
+    });
+
+    afterEach(() => {
+      portalConfig.environment = originalEnvironment;
+    });
+
+    it('returns an empty set outside production', async () => {
```
Kroustille: An hidden service definition needs to be created to assert it's not returned by `ServiceDefinitionDomain.loadHiddenIdentifiers`

apps/backend/src/modules/service/definition/service-definition.resolver.ts
```diff
@@ -1,13 +1,23 @@
...
+import { ServiceDefinitionDomain } from './service-definition.domain';
 
 const resolvers: Resolvers = {
   ServiceDefinition: {
     service_capability: ({ id }, _) =>
       ServiceCapabilityDomain.loadServiceCapabilitiesBy({
         service_definition_id: id,
       } as ServiceCapabilityMutator),
+    hidden: ({ hidden }) =>
+      portalConfig.environment === 'production' && !!hidden,
```
Kroustille: Maybe we could add an util function 'isProduction' which just returns `portalConfig.environment === 'production'`. What do you think ?

apps/backend/src/modules/service/definition/service-definition.graphql
```diff
@@ -16,6 +16,11 @@ type ServiceDefinition implements Node {
   name: String!
   description: String
   public: Boolean
+  hidden: Boolean
   identifier: ServiceDefinitionIdentifier!
   service_capability: [ServiceCapability]
 }
+
+type Query {
+  hiddenServiceDefinitionIdentifiers: [ServiceDefinitionIdentifier!]!
```
Kroustille: I'm wondering if we should invert behavior: with this query, any unauthenticated user can retrieve the hidden service definition identifiers.

I'm not sure it has a big impact on security but it's something I wanted to notice.

Another future improvement would be to have a route dedicated to returning the data needed for public and private navigation, so the entire hidden/visible behavior stays on the server and client only makes one request to fetch every data needed.

Kroustille (commented):
Should we also filter services from sitemap when its service definition is hidden ?

apps/backend/src/modules/service/instance/service-instance.app.ts
```diff
@@ -44,6 +47,9 @@ export const ServiceInstanceApp = {
     let subscription = existingSubscription;
 
     if (!subscription) {
+      if (!service.public) {
+        throw ForbiddenAccess(ErrorCode.ServiceInstanceNotPublic);
```
Kroustille:
```suggestion
        throw new Error(ErrorCode.ServiceInstanceNotPublic);
```

The ErrorCode is then mapped to a ForbiddenAccess at the resolver layer.

Kroustille (commented):
`serviceInstanceById` query from the service instance resolver is not protected as the `serviceInstanceByIdAndGrantAccess` query. It's means any authenticated user with a private service instance id can access this service instance data.

The same happens with `DocumentApp.loadDocuments`: anyone with a private service instance id can access documents from this service instance.

Kroustille (approved):

PR #3046, opened 2026-08-06
feat(seo): allow admins to update meta title and meta descriptions for services libraries (#3040)

hervyt:
# Context
> To improve SEO, the user is now able to edit metadata (title and description) for a service and reuse these data in the SEO. If we don't have existing data, the fallback metadata is the original data (serviceName for public/XTM Hub for private). 

## How to test
- On Google Chrome, get the extension json-ld viewer 
- Go on a service library on public page, you should see by default in the meta-title and meta-desc the title and name of the service.  (with JSON-LD Viewer)
- Connect as a user ADMIN or from Marketing group 
- Go on a library service 
- Click on the edit button 
- Fill all the fields (meta title and desc for each locale) and validate 
- Logout 
- Go to the public page of this service 
- You should now see the new values (if you change language in the URL, the values change for the correct locale)  (with JSON-LD Viewer)
- Login
- Go on the private page of this service 
- - You should now see the new meta values (with JSON-LD Viewer) 
- The edit button should now be available if you don't have correct capabilities

https://github.com/user-attachments/assets/8d3eef56-ab32-40b8-8e7a-cd958424f995

## Related issues

- Related to #3040

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [X] Integration tests
- [ ] E2E tests
- [X] Local manual tests


apps/frontend/src/components/service/service.graphql.ts
```diff
@@ -28,6 +28,38 @@ export const UpdatePlatformServiceMetadata = graphql`
   }
 `;
 
+export const EditSeoServiceInstanceMutation = graphql`
```
Kroustille: We should take the opportunity to use `react-query` for this newly added mutation/query instead of relay

apps/frontend/src/components/service/document/ui/LibraryUpdateMetadatas.tsx
```diff
@@ -0,0 +1,227 @@
...
+    {
+      label: string;
+      inputProps: {
+        maxLength: number;
+      };
+    }
+  >
+);
+
+export const LibraryUpdateMetadatas = () => {
```
Kroustille: Metadata is invariant and can be never be plural


```suggestion
export const LibraryUpdateMetadata = () => {
```

apps/frontend/app/(public)/[locale]/seo-service-instance.graphql.ts
```diff
@@ -34,3 +34,20 @@ export const SeoServiceInstanceQuery = graphql`
     }
   }
 `;
+
+export const SeoServiceInstanceMetadataQuery = graphql`
```
Comment (Kroustille) [inline apps/frontend/app/(public)/[locale]/seo-service-instance.graphql.ts:38]: Same here for the react-query/relay

apps/frontend/app/(application)/app/layout.tsx
```diff
@@ -14,18 +17,62 @@ import { loadMeUser } from '@/utils/load-me-user';
...
+const extractServiceInstanceId = (pathname: string): string | null => {
+  const decodedPathname = decodeURIComponent(pathname.split('?')[0] ?? '');
+  if (!decodedPathname.includes('/service/')) {
+    return null;
+  }
+  const match = decodedPathname.match(regExpUUID);
+  return match?.[0] ?? null;
+};
+
 export const generateMetadata = async (): Promise<Metadata> => {
```
Kroustille, line 24: I think we should not update the generateMetadata with an id extraction from URL but used a new generateServiceMetadata function to compute values and use it on service pages (exported by generateMetadata)

apps/backend/knexfile.ts
```diff
@@ -85,6 +86,7 @@ type BaseDatabaseType =
   | 'DeploymentRequestQuota'
   | 'ServiceGroup'
   | 'ServiceGroup_User'
+  | 'SEO_ServiceInstance'
```
Kroustille, line 88: This value was added twice

apps/frontend/app/(public)/[locale]/cybersecurity-solutions/[slug]/page.tsx
```diff
@@ -66,20 +80,20 @@ export async function generateMetadata({
...
-  const description = localizedCardDescription(serviceInstance, t);
 
   return buildSeoPageMetadata({
     baseUrl,
     locale,
     pathname,
-    title: `${name} | XTM Hub`,
-    description,
-    imageAlt: name,
+    title: seoMetadata?.meta_title || serviceInstance.name,
```
Comment (Kroustille) [inline apps/frontend/app/(public)/[locale]/cybersecurity-solutions/[slug]/page.tsx]: Is it normal we dropped the `| XTM Hub` from title ?

apps/backend/src/modules/service/instance/seo-service-instance/seo-service-instance.domain.ts
```diff
@@ -0,0 +1,41 @@
...
+        language,
+        ...payload,
+      })
+      .onConflict(['service_instance_id', 'language'])
+      .merge(payload)
+      .returning('*');
+
+    if (!seoServiceInstance) {
+      throw new Error(NotFoundErrorCode.SeoServiceInstanceNotFound);
+    }
```
Kroustille, line 37: This case can never happen because and `insert/returning` is done, meaning that we're sure to have a `seoServiceInstance` when no Error is thrown

apps/backend/src/modules/service/instance/seo-service-instance/seo-service-instance.app.ts
```diff
@@ -0,0 +1,38 @@
...
+    language,
+  }: QuerySeoServiceInstanceMetadataArgs): Promise<
+    SeoServiceInstanceMetadata[]
+  > => {
+    const field: SEOServiceInstanceMutator = {
+      service_instance_id: service_instance_id as ServiceInstanceId,
+    };
+    if (language != null) {
+      field.language = language as SEOServiceInstanceLanguage;
+    }
```
Kroustille, line 21:
```suggestion
    const field: SEOServiceInstanceMutator = {
      service_instance_id: service_instance_id,
    };
    if (language != null) {
      field.language = language;
    }
```

Input fields are already typed by GraphQL

apps/frontend/src/components/service/document/ui/LibraryUpdateMetadata.tsx
```diff
@@ -0,0 +1,227 @@
...
+    acc[descriptionField] = '';
+    return acc;
+  },
+  {} as LibraryUpdateFormValues
+);
+
+const libraryUpdateFieldConfig = SEO_LOCALES.reduce(
+  (acc, { titleField, descriptionField, label }) => {
+    acc[titleField] = {
+      label: `Meta-title (${label})`,
```
Kroustille: Shouldn't we translate Meta-title and Meta-description here ?

apps/backend/src/modules/service/instance/seo-service-instance/seo-service-instance.domain.ts
hervyt: The if is a type-safety guard: even though INSERT ... RETURNING should yield a row at runtime, Knex still types the result as a plain array, so TypeScript cannot guarantee the first element exists. This check makes that assumption explicit and keeps the function safely typed.

apps/frontend/src/components/service/document/ui/LibraryUpdateMetadata.tsx
hervyt: I don't understand your comment here... Each field is distributed to its locale (we're in the admin form, not in the usage of these meta-title and meta-description (that are correctly translated on the correct page ;) )

apps/backend/src/modules/service/instance/seo-service-instance/seo-service-instance.app.ts
```diff
@@ -0,0 +1,37 @@
...
+    service_instance_id,
+    language,
+  }: QuerySeoServiceInstanceMetadataArgs): Promise<
+    SeoServiceInstanceMetadata[]
+  > => {
+    const field: SEOServiceInstanceMutator = {
+      service_instance_id: service_instance_id,
+    };
+    if (language != null) {
+      field.language = language as SEOServiceInstanceLanguage;
```
Kroustille: I think that this cast should be removed by updating kanel config to make sure both GraphQL and kanel enums are the same type.

apps/frontend/src/components/service/document/ui/LibraryUpdateMetadata.tsx
Kroustille: Yes but the `Meta-title` or `Meta-descriptions` labels itself are not translated for the user who wants to fill it. It's kinda nitpicking and may be not mandatory for this PR 😉

hervyt: AHHHHHH! Understood 🤣

Kroustille (approved):

PR #3062, opened 2026-08-10
feat(integration): ingest license_type, contact and solution categories for legacy connectors (#2997)

Antoine-Quintin:
# Context
>

For legacy (pre-decoupling) connectors, ingest three new fields from the OpenCTI
connectors manifest:

- `license_type` (`Free`/`Commercial`) → `Document_Metadata`
- `contact` → `Document_Metadata`
- `solution_categories` → `Object_SolutionCategory`

The ingestion was replayed on the feature environment (`dev-pr-3062`,
tag `7.260803.0`) — the before/after below shows the result on the same
connector:

**Before ingestion** (fresh feature-env DB):
image

**After ingestion:**
image

To replay it yourself (here or locally):

```graphql
query UpdateOpenCTIManifest($tag: String!) {
  updateOpenCTIManifest(tag: $tag) { success }
}
```
variables `{ "tag": "7.260803.0" }`, header `x-xtm-hub-token`.

The query is fire-and-forget: `success: true` only acknowledges the
request — check the data (UI or DB) after a moment, not the GraphQL response.

## Related issues

- Related to #2997

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -61,6 +61,7 @@ export type DocumentData<
 > = Omit<Partial<T>, 'use_cases'> & {
   use_cases?: TUseCase[];
   solution_category?: TSolutionCategory;
+  solution_categories?: string[];
```
Kroustille, line 72:
```suggestion
  solution_categories?: TSolutionCategory[];
```

??

Kroustille, line 72: Would it be better to have only one property array (with possibly only one element) instead of two representing the same things ?

apps/backend/src/modules/shareable-resource/opencti/integration/ingest-manifest/ingest-manifest.helper.test.ts
```diff
@@ -14,6 +25,39 @@ import sampleManifest from './test/sample-manifest.json';
...
 describe('ingest manifest helper', () => {
   describe('extractManifestInfo', () => {
+    let warnSpy: MockInstance;
+
+    beforeEach(() => {
+      warnSpy = vi.spyOn(logApp, 'warn').mockImplementation(() => {});
+    });
+
+    afterEach(() => {
+      warnSpy.mockRestore();
```
Kroustille: This restore is not required since we're doing a `restoreAllMocks`. I need to add an ESLint rule for that 😛

apps/backend/src/modules/document/document.app.ts
```diff
@@ -812,6 +813,19 @@ const upsertDocument = async <T extends DocumentModel>(
...
+    if (documentData.solution_categories !== undefined) {
+      if (documentWasUpdated) {
+        await objectSolutionCategoryDomain.deleteObjectSolutionCategoryBy({
+          object_id: toObjectSolutionCategoryObjectId(document.id),
+        });
+      }
+      await solutionCategoryApp.linkSolutionCategoriesByNameToObject({
+        objectId: toObjectSolutionCategoryObjectId(document.id),
+        names: documentData.solution_categories,
+        product: PlatformIdentifier.Opencti,
```
Kroustille: We could use the mapping with product to make sure it continues to work when we add an upsert with a different product

apps/backend/src/modules/document/document.app.test.ts
```diff
@@ -1464,6 +1500,234 @@ describe('documentApp', () => {
         });
       });
     });
+
+    describe('solution_categories linking', () => {
```
Kroustille: We could move this tests in a new `document.app.solution-categories.test.ts` to prevent `document.app.test.ts` from growing more

apps/backend/src/modules/document/domain/document.domain.ts
Antoine-Quintin: The two fields intentionally carry different things. solution_category is a category id (TSolutionCategory defaults to SolutionCategoryId), used by the form path which inserts it directly as the FK. solution_categories carries names straight from the manifest — the upstream key is literally solution_categories and ships names, verified on tag 7.260803.0 — resolved at link time via linkSolutionCategoriesByNameToObject. Both ingestion paths (v1 manifest and v2 fragments) speak names; only the form path speaks ids.

Typing the plural as TSolutionCategory[] would make it SolutionCategoryId[] on the ingestion path (which instantiates DocumentData<T, string>, so the third generic takes its default), and the zod-extracted strings can't satisfy the branded id type without a cast.

apps/backend/src/modules/document/document.app.ts
Antoine-Quintin: Agreed — this is the latent trap : a non-OpenCTI caller setting solution_categories would silently resolve against the wrong referential. Two options I see: derive the product from the document's service instance as you suggest, or thread a FiligranProduct-typed parameter through the call — which would also fix the current string coupling between PlatformIdentifier, FiligranProduct and the free-form manifest platform field.

apps/backend/src/modules/document/domain/document.domain.ts
Kroustille: I understand why we have `strings` and `TSolutionCategory` and I think `DocumentData` needs to be changed entirely to be more explicit on its purpose between document app and domain, but it's not in the scope of this PR.

On an other side there is still an incoherence because with connectors manifest ingestion, we're creating multiple solution categories for a document but the GraphQL typing only mention a single `solution_category` and the dataloader will load multiple categories.

GraphQL typing needs to be updated to `solution_categories: [SolutionCategory!]` in integration and document graphql files.

apps/backend/src/modules/document/document.app.ts
Kroustille: I think the second solution is the right one 😁

apps/backend/src/modules/document/domain/document.domain.ts
Antoine-Quintin: Done — the GraphQL output is now solution_categories: [SolutionCategory!] across the integration types, the dataloader accumulates (which also removes the non-deterministic single-category pick), and the frontend consumers are updated. The form input stays single-select pending the product decision — the shared field takes the first ingested category as its default.

apps/frontend/src/components/service/form/SolutionCategoryField.tsx
```diff
@@ -30,11 +30,15 @@ export const ServiceFormSolutionCategoryField = ({
 }: ServiceFormSolutionCategoryFieldProps) => {
   const t = useTranslations();
   const categories = useSolutionCategories(product);
-  const selectedCategoryId = field.value ?? document?.solution_category?.id;
+  // The form input is still single-select while the API now returns a list:
+  // take the first ingested category as the default, pending the product
+  // decision on form cardinality.
```
Kroustille, line 35: For the product decision, is there a created ticket or opened question somewhere ? To make sure this comment will not get lost in code.

Kroustille (approved):

apps/frontend/src/components/service/form/SolutionCategoryField.tsx
Antoine-Quintin: not for now, I'll create it after a talk with product team

hervyt (approved):

Issue #2863, opened 2026-07-20
feat(admin): Ability for admin platform to delete users


EllynBsc:
## Use case
Ability for admin platform to delete users. Please add a basic confirmation modal to make sure the admin user knows that he/she is removing someone from the Hub.

Technical note:
To be able to delete an User we need to delete the organization beforehand as there is a constraint between user and organization.

User must be hard deleted to make sure there's no remaining data concerning them in our DB

jbanety:
### 🚧 Constraints

- No `deleteUser` mutation exists yet in the GraphQL schema/resolvers. 
However, an unused helper  `removeUser`  already exists in `user.helper.ts:351-367,  it deletes the  `User`  row then deletes the personal-space  `Organization`  (same  `id`  as the user, per convention). It's dead code today (not called anywhere). 
Verify it's still correct/safe before wiring it up, don't assume it works as-is.
- FK reality check:  `User.selected_organization_id`  →  `Organization.id`  is a plain FK with no cascade. 
This blocks deleting the personal-space `Organization` while a  `User`  still points to it.
So the correct order is actually delete `User` first, then its personal-space `Organization` (matches what  `removeUser`  already does). Double-check the technical note's phrasing against this, if "delete org before user" was intended literally, it contradicts the existing FK direction; confirm with the ticket author which specific constraint they hit.
- Multiple other non-cascading FKs reference  `User`  without  `onDelete: CASCADE `:  `service_instance_creation.subscriber_id` ,  `transfer_perso_space.from_user_id/to_user_id` ,  `deploymentRequest` ,  `group_groupUser`  (see migrations grep). 
Deleting a user with any live rows in these tables will fail with a FK violation, this is the real risk area, not just the `Organization` constraint. 
Dev must inventory all FK references to  `User`  before implementing and decide per-table: block deletion with a clear error or cascade/nullify.
- `User_Organization` ,  `UserOrganization_Capability` ,  `UserService` , malware-analysis, pending-org tables do cascade on user delete. 
No special handling needed there.

### 🐷 Requirements

- Add  `deleteUser(id: ID!)`  mutation,  `@auth(portalCapa: [BYPASS]`)  (platform-admin only, consistent with  `deleteOrganization` ).
- Confirmation modal before delete (basic, "Are you sure you want to remove this user from the Hub?", same UX pattern as existing delete/remove confirmations elsewhere in admin UI).
[...]

Issue #11, opened 2024-09-18
Alignment on UI look


EllynBsc:
**Detailed Description**
This issue aims to address the inconsistencies in UI design alignment between the OCTI and OBAS components within our project. The current misalignment may lead to a less intuitive user experience and inconsistencies throughout the platform. Our task is to ensure our design system aligns with both components for a seamless user interface. For more details regarding the specific inconsistencies, please refer to the following issue: [Issue#13](https://github.com/FiligranHQ/XTMHub/issues/13).

PR #3066, opened 2026-08-10
feat(user): add ability to delete an user and handle edge cases (#2863)

Kroustille:
# Context
>

| # | Link | Setup | UI action (manual) |
|---|---|---|---|
| 1 | `User_Organization.user_id` (CASCADE) | Create a test user. Optionally also add them as a member of a shared organization. | In Admin → Users, delete the test user's row. If they belong to a shared org, first check **Admin → Organizations → [org] → Members** (or **Manage → Organization** as that org's admin) to see them listed; after deletion, reload that members list and confirm they're gone. |
| 2 | `UserOrganization_Capability` (cascades via #1) | Create a test user who has been given a special permission ("capability") inside a shared organization. | As an org admin, open **Manage → Organization → Users**, grant the target user a capability (e.g. "Manage Access") via the row's capability picker; as super-admin, delete that user from Admin → Users; re-open the org's user list and confirm the row (and its capability badge) is gone. |
| 3 | `User_Organization_Pending.user_id` (CASCADE) | Create a test user who has asked to join a different organization, and that request hasn't been approved yet. | Have the pending user request to join a shared org (join-request UI), leaving it pending in **Manage → Organization → Pending users**; as super-admin, delete that user from Admin → Users; reload the org's pending-users tab and confirm the request row is gone. |
| 4 | `User_RolePortal.user_id` (CASCADE) | Create a test user who has been assigned a special platform-wide role. | **No manual UI path** — these roles come automatically from single-sign-on group settings on login, not from any admin screen. The only manual check available is: log the user in via SSO so they get the role, then confirm the delete still works without error (nothing to visually confirm beyond that). |
| 5 | `User_Service.user_id` (CASCADE) | Create a test user who has been added to a subscription for a service. | As an org admin with an active subscription, go to **Manage → Service → [instance] → Subscription**, add the target user to the subscription's user list; as super-admin, delete that user from Admin → Users; reload the subscription's user list and confirm the row disappeared. |
| 6 | `UserService_Capability` (cascades via #5) | Create a test user who has been added to a subscription and given a special permission on it. | Same subscription screen as #5: grant the user a service-level capability (e.g. via the capability selector on their subscription-user row), then delete the user from Admin → Users and confirm both the row and its capability badge are gone on reload. |
| 7 | `ServiceGroup_User.user_id` (CASCADE) | Create a test user who belongs to a group of users for a specific service. | For a trial-type service, open **Manage → Service → [trial instance] → Manage users**, add the user to a group and save; as super-admin, delete the user from Admin → Users; reopen the manage-users screen and confirm they no longer appear in that group's list. |
| 8 | `User_TransferRequest.from_user_id` / `to_user_id` (blocked) | Create two test users and start (but don't finish) a request to transfer one user's personal space to the other. | Initiate a personal-space transfer from the user's **Profile → Transfer personal space** flow, leave it pending (don't accept/decline); as super-admin, try to delete either user from Admin → Users — it should be blocked with a clear warning message. |
| 9 | `DeploymentRequest.user_requester_id` (blocked) | Create a test user who has submitted a request to deploy a service. | As that user, submit a deployment request (e.g. via a service's "Request deployment" action in Manage → Service); as super-admin, try to delete that user from Admin → Users for each stage of that request's lifecycle — it should always be blocked with a clear warning message. |
| 10 | `DeploymentRequest.cancellation_user_id` (blocked) | Create two test users: one submits a deployment request, the other cancels it. | As the second user (an admin with cancel rights), cancel the first user's pending deployment request from the deployment-requests admin view; as super-admin, try to delete the second user from Admin → Users — it should be blocked with a clear warning message. |
| 11 | `Document.uploader_id` (reassigned) | Create a test user who has uploaded a document. | As that user, upload a document via the relevant service's document/vault screen; as super-admin, delete that user from Admin → Users (should succeed, no warning); reopen the document and confirm the "uploaded by" field now shows a system/platform account instead of erroring or showing a blank reference. |
| 12 | `Document.remover_id` / `updater_id` (reassigned) | Create a test user who has edited, deleted, or restored a document that someone else uploaded. | As that user, edit/replace or delete-then-restore a document uploaded by someone else; as super-admin, delete that user; confirm the document's history/metadata view (if shown) doesn't break and attributes the action to a system account. |
| 13 | `Epic.uploader_id` / `updater_id` (reassigned) | Create a test user who has created or updated a roadmap item ("Epic"). | As that user, create/update a roadmap item via the relevant admin/roadmap screen; delete that user from Admin → Users; reopen the roadmap item and confirm it still renders (attributed to a system account) rather than showing a broken reference. |
| 14 | `PlatformConfiguration.registerer_id` (blocked) | Create a test user who has registered a service instance, both while it's active and while it's inactive. | As that user, register a service instance (Manage → Service → register flow) for both an active and an inactive case; as super-admin, try deleting that user from Admin → Users in each case — it should always be blocked with a clear warning message. |
| 15 | `OneClickDeployment.user_id` (SET NULL) | Create a test user who has triggered a "one-click deployment" through the embed widget. | As that user, trigger a one-click deployment (embed widget flow); as super-admin, delete that user from Admin → Users — deletion succeeds. The user one click deploy record is kept but the user is set to null. The deployed resources displays a deleted user info on homepage |
| 16 | `Document.uploader_organization_id` → `Organization` (no FK, out of scope) | Not tested — intentionally out of scope. Would involve a document uploaded by one user but credited to another user's organization. | Same vault/document screen as #11/#12 — after deleting the organization's owning user, the document's "organization" attribution may silently point at a deleted org; not currently checked or fixed (explicitly out of scope). |
| 17 | `User.selected_organization_id` → `Organization` (reverse link, ordering) | Create a test user (their personal space is automatically their own organization). | Delete the user from Admin → Users and confirm no error message appears (an ordering bug would otherwise surface as a generic error). |
| 18 | `sessions` (no FK, explicit cleanup) | Create a test user and log them in so they have an active session. | Log in as that user in one browser/incognito session; in another session (as super-admin), delete that user from Admin → Users; go back to the first browser and refresh — it should log the user out rather than keep working. |
| 19 | Last-organization-member guard | Create a shared organization with only one member: the test user. | As super-admin, try to delete that user from Admin → Users — it should be blocked with a clear warning message. |
| 20 | Personal-space pending-users guard | Create two test users: one has asked to join the other's personal space, and that request hasn't been approved yet. | As super-admin, try to delete the user whose personal space has the pending request — it should be blocked with a clear warning message. |

## Related issues

- Related to #2863

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/src/migrations/20260810120000_user_reference_integrity.js
```diff
@@ -0,0 +1,81 @@
...
+    );
+    const danglingCount = Number(count);
+
+    if (danglingCount === 0) {
+      continue;
+    }
+
+    if (!systemUser) {
+      throw new Error(
+        `Cannot backfill ${danglingCount} dangling "${reference.table}"."${reference.column}" reference(s): system user ${SYSTEM_USER_UUID} does not exist`
```
carinelebas: question: I wonder if we really want a migration to throw 🤔 If it fails, it will prevent server start, even in production?

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -171,6 +172,24 @@ export const DeploymentRequestDomain = {
...
 
+  countDeploymentRequestsForUser: async (userId: UserId): Promise<number> => {
+    const result = await db<DeploymentRequest>('DeploymentRequest')
+      .where('user_requester_id', '=', userId)
+      .count<[{ count: string }]>('id as count')
+      .first();
+    return Number(result?.count ?? 0);
+  },
+
+  countDeploymentRequestsCancelledByUser: async (
```
carinelebas: This is the same query as above, just with a different where condition. Could we use the same function ?

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -65,6 +69,23 @@ export type DocumentData<
 };
 
 export const DocumentDomain = {
+  reassignUserDocumentsToSystemUser: async (userId: UserId): Promise<void> => {
+    await db<DocumentModel>('Document')
```
carinelebas: question: should this be done in a transaction to have a all or nothing behavior?

apps/backend/src/modules/organization-management/organization/organization-membership.util.ts
```diff
@@ -0,0 +1,42 @@
+import { OrganizationId } from '../../../model/kanel/public/Organization';
+import { UserOrganizationDomain } from '../user/user-organization/user-organization.domain';
+import { UserOrganizationPendingDomain } from '../user/user-pending/user-organization-pending.domain';
+
+/**
+ * Shared invariant checks reused by both the "delete this organization"
+ * (OrganizationApp) and "would removing this user break this organization"
+ * (UserAdminApp) flows, so the two paths can't drift on the underlying
```
carinelebas: I think it's better to avoid referencing the calling functions in the comments. It usually becomes obsolete very fast because it's not maintained, for example at the first refactor.

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -65,6 +69,23 @@ export type DocumentData<
...
+      .where('uploader_id', '=', userId)
+      .update({
+        uploader_id: SYSTEM_USER_UUID,
+        uploader_organization_id: PLATFORM_ORGANIZATION_UUID,
```
theochhean: 🤔 je pensais pareil que le commentaire copilot, c'est quoi l'explication du coup de tout passer en PLATFORM_ORGANIZATION_UUID ?

apps/backend/src/modules/organization-management/organization/organization.app.test.ts
```diff
@@ -110,24 +112,176 @@ describe('organizationApp', () => {
...
+      await TestHelper.organization.delete({ id: organization.id });
+      await UserHelper.removeUser({ id: linkedUser.id });
+    });
+
+    it('should not apply the guards on a personal space organization', async () => {
+      const user = await UserHelper.createUserWithPersonalSpace(
+        { email: `delete-orga-${uuidv4()}@delete-orga-test.io` },
+        { sendWelcomeEmail: false }
+      );
+      const personalSpaceId = user.id as unknown as OrganizationId;
```
theochhean: je vois ce cast un peu partout dans le code, je comprend que c'est pour la transition UserId -> OrganizationId, mais je pense qu'on pourrait le factoriser dans un utils par exemple pour qu'il soit mieux documenté et maintenu

apps/backend/src/modules/organization-management/user/user-admin/user.admin.app.ts
```diff
@@ -8,23 +8,38 @@ import {
...
+import { DocumentDomain } from '../../../document/domain/document.domain';
+import { EpicDomain } from '../../../xtm-platform-roadmap/epic.domain';
 import { OrganizationDomain } from '../../organization/organization.domain';
 import { UserDomain } from '../user-domain/user.domain';
 import { UserOrganizationDomain } from '../user-organization/user-organization.domain';
 import { UserOrganizationPendingDomain } from '../user-pending/user-organization-pending.domain';
 import { UserHelper } from '../user.helper';
+import { UserAdminGuard } from './user.admin.guard';
+
+const FOREIGN_KEY_VIOLATION_CODE = '23503';
```
theochhean, line 38: pour ma compréhension ça correspond à quoi ce code ?

apps/backend/src/modules/organization-management/user/user-admin/user.admin.guard.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  },
+
+  assertUserIsNotLastOrganizationAdministrator: async (userId: UserId) => {
+    const sharedOrganizations = await loadSharedOrganizations(userId);
+
+    for (const organization of sharedOrganizations) {
+      if (await isUserLastOrganizationAdministrator(userId, organization.id)) {
+        throw new Error(ErrorCode.CantRemoveLastAdministrator);
```
theochhean, line 78: J'ai l'impression que ce code est manquant dans le mapping front DELETION_BLOCKED_ERROR_CODES

apps/frontend/src/components/admin/user/DeleteUser.tsx
```diff
@@ -0,0 +1,88 @@
...
+          onClose={() => {
+            setBlockedReasonKey(null);
+            setOpen(false);
+          }}
+          title={t('UserActions.DeletionBlocked')}
+          description={t(blockedReasonKey)}>
+          {null}
+        </DialogInformative>
+      )}
+    </>
```
theochhean: pas trop fan du fragment en wrapper j'aurai plutôt utilié un ternary ici

apps/frontend/src/components/admin/user/UserList.tsx
```diff
@@ -407,6 +449,9 @@ const UserList = ({ organization }: UserListProps) => {
           pagination,
           columnOrder,
           columnVisibility,
+          columnPinning: {
+            right: ['actions'],
+          },
```
theochhean, line 455: si j'ai bien compris `actions` n'est présent que si `isAdminPath && canDeleteUser` est-ce qu'il faudrait ajouter ce check ici aussi ?

```suggestion
          columnPinning: {
            right: isAdminPath && canDeleteUser ? ['actions'] : [],
          },
```

apps/backend/src/modules/organization-management/user/user-admin/user.admin.app.ts
Kroustille: It's PostgreSQL's error code for `foreign_key_violation` (class 23 Integrity Constraint Violation). See the full list at https://www.postgresql.org/docs/current/errcodes-appendix.html. 

Added a comment above the constant for clarity.

apps/backend/src/modules/document/domain/document.domain.ts
Kroustille: I want to keep coherence between the `uploader_id` which becomes the system user id, and the organization where he has rights (Filigran). 

For the moment, we only have documents created by Filigran users so it's not really a problem but we will have to think about this again when we open documents creation for users from other organizations

apps/backend/src/migrations/20260810120000_user_reference_integrity.js
```diff
@@ -39,13 +39,30 @@ export async function up(knex) {
     table.dropForeign(['user_requester_id']);
     table.foreign('user_requester_id').references('id').inTable('User');
   });
+
+  await knex('OneClickDeployment')
```
carinelebas, line 43: question: is it really something that we want to do ? we are using one click deployment tables to display the number of resources deployed (overall, and on a specific platform). If we remove a user, it doesn't mean the resource was not deployed?

carinelebas (approved):

Issue #3063, opened 2026-08-10
fix(SSE): Too many SSE connections opened per tab


carinelebas:
SSE subsription is limited to 6 per browser.  

bug: `NotificationButton`  (pending-user badge) is mounted twice in  Header.tsx  (desktop  mobile:hidden  block + mobile  sm:hidden  block, both always in the DOM), each independently subscribing to the same  pendingUserList  subscription.  -> duplicate subscription for pending users

Improvement : The  graphql-sse   client can be created with  singleConnection: true, which would multiplex all subscriptions over one SSE connection per tab.

PR #3070, opened 2026-08-11
fix(sse): sse singleConnectio, Header NotificationButton single mount (#3063)

theochhean:
# Context
>

SSE subsription is limited to 6 per browser.  

bug: NotificationButton  (pending-user badge) is mounted twice in  Header.tsx  (desktop  mobile:hidden  block + mobile  sm:hidden  block, both always in the DOM), each independently subscribing to the same  pendingUserList  subscription. -> duplicate subscription for pending users

What has been done : 
- Set `singleConnection` so that a tab would only have one connection and not multiple ones
- Refactor `NotificationButton` to only call it once

before changes : a single page could do multiple SSE connections, resulting in reaching the 6 connection limit early and therefore opening new tabs would load indefinitely/not load everything
- Open a tab logged in in an organization with AdministrateOrganization or ManageAccess rights
- Open other tabs the same way
- Limit will reach early, page will not fully load / not at all (for me at some point on one tab, XTM Platform Roadmap on the Left Menu does not appear, opening a new tab will result in an indefinite loading)

after changes : only 1 connection per tab, limit is still to 6 but that means we can open up to 6 tabs before reaching the limit
- Open as many tabs the same way, limit will always reach at the 6th one and not early

## Related issues

- Related to #3063

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/frontend/src/components/Header.tsx
+++ b/apps/frontend/src/components/Header.tsx
@@ -26,7 +26,7 @@ import Link from 'next/link';
 
 import { OrganizationCapability } from '@graphql/generated';
 import { usePathname, useRouter } from 'next/navigation';
-import { useContext, useEffect, useState } from 'react';
+import { useCallback, useContext, useEffect, useState } from 'react';
 import { useMutation } from 'react-relay';
 
 // Component interface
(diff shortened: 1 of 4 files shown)


apps/frontend/src/components/Header.tsx
```diff
@@ -66,99 +66,102 @@ const HeaderComponent = ({ displayLogo }: HeaderComponentProps) => {
...
+                      commitLogoutMutation({
+                        variables: {},
+                        updater: (store) => {
+                          store.invalidateStore();
+                        },
+                        onCompleted() {
+                          router.push('/');
+                          router.refresh();
+                        },
+                      });
```
Kroustille: Not mandatory improvement: this logout logic is duplicated in this file and can be mutualize.

Kroustille (approved):

Issue #3079, opened 2026-08-12
feat(solution_categories): Allow the user to fill a shareable resource with multiple solution_categories


hervyt:
## Use case

Today, an admin user that is creating a shareable resource on XTM Hub can upload only 1 solution_category. However, connectors from the manifest can have mutliple solution categories. 

## Proposed solution

To be more coherent, allow the user to create a shareable resource with multiple solution_categories. 

Change in the frontend the <Select> to a <multiSelect>

Handle this in the backend.

PR #3082, opened 2026-08-13
feat(integration): Make many solution_categories possible (#3079)

hervyt:
# Context
> Add the possibilty to the user to add multiple solution_categories to an integration

## How to test
- Add an integration, you should be able to add with multiple solutionCategories 
- Read this integration with all its solutionCategories 
- Edit an integration with multiple solutionCategories

https://github.com/user-attachments/assets/1ca0cd02-f357-46ae-bd3b-98f322f5eb52

## Related issues

- Related to #3079 

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [X] E2E tests
- [X] Local manual tests

# Additional information


apps/backend/src/modules/document/document.app.ts
```diff
@@ -208,11 +209,13 @@ export const DocumentApp = {
...
-      if (documentData.solution_category) {
-        await objectSolutionCategoryDomain.insertObjectSolutionCategory({
-          object_id: toObjectSolutionCategoryObjectId(document.id),
-          solution_category_id: documentData.solution_category,
-        });
+      if (documentData.solution_categories?.length) {
+        for (const solutionCategoryId of documentData.solution_categories) {
+          await objectSolutionCategoryDomain.insertObjectSolutionCategory({
+            object_id: toObjectSolutionCategoryObjectId(document.id),
+            solution_category_id: solutionCategoryId as SolutionCategoryId,
```
Kroustille: I'm wondering if it's the right way to handle it since the `solution_categories` in documentData is considered as possibly the category name when it comes from manifest ingestion, or id when it's coming from the UI form 🤔 

We could use `solution_category_ids` to specify it's the ids with property typing for the form, and `solution_category_names` when it's provided from the manifest ingestion.

This way, we're sure to know which type of data we're handling.

What do you think ? 

Another solution would be to separate solution categories from documentData and use it as an additional parameter in document app functions.

apps/backend/src/modules/document/document.app.ts
```diff
@@ -395,15 +398,16 @@ export const DocumentApp = {
...
-        await objectSolutionCategoryDomain.insertObjectSolutionCategory({
-          object_id: toObjectSolutionCategoryObjectId(parentDocumentId),
-          solution_category_id: input.solution_category,
-        });
+        for (const solutionCategoryId of input.solution_categories) {
+          await objectSolutionCategoryDomain.insertObjectSolutionCategory({
+            object_id: toObjectSolutionCategoryObjectId(parentDocumentId),
+            solution_category_id: solutionCategoryId,
+          });
+        }
```
Kroustille: If you do the update by separating solution category properties, this part of code is duplicated with the create one and could be refactored.

Kroustille (commented):
There is a remaining `solution_category` property in DocumentData that can be removed

Kroustille (approved):

PR #3113, opened 2026-08-18
feat(deployment): cancellation and expiry cascade for bundles (#3096)

Antoine-Quintin:
# Context
>

Cancelling or expiring a bundle now cascades to all its products, and a single product of a bundle can no longer be cancelled.

Three behaviours:

Cancellation cascade — cancelling a bundle cancels the bundle and every product, in a single transaction. counts_in_orga_quota and target_state are still computed per row, the cancellation reason is written on every row, and one telemetry event is emitted per row.

Expiry cascade — loadTrialsToExpire now selects bundles and standalone trials, and excludes bundle children (they expire with their parent). Same all-or-nothing transaction.

Single product cancellation refused — a row with a parent_id is rejected with CANT_CANCEL_BUNDLE_PRODUCT, for customers and platform admins alike.

Bruno (bruno/deployment/) and psql.

CreateDeploymentRequest with type: bundle, products: ["xtmone", "opencti"]
CancelDeploymentRequest on a child → CANT_CANCEL_BUNDLE_PRODUCT
CancelDeploymentRequest on the bundle → bundle + both children cancelled, reason on every row
Expiry: set hub_status = 'active', actual_state = 'active', end_date = now() - interval '1 day' on the family, then trigger expireTrials (temporarily set the cron to '* * * * *') → all three rows expired / removed

## Related issues

- Related to #3096

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -373,7 +373,9 @@ export const DeploymentApp = {
       throw new Error(NotFoundErrorCode.DeploymentRequestNotFound);
     }
 
-    const previousHubStatus = deploymentRequest.hub_status;
+    if (deploymentRequest.parent_id) {
+      throw new Error(ForbiddenErrorCode.CantCancelBundleProduct);
+    }
```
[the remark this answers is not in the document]
Antoine-Quintin: refusing a bundle child is an unconditional business rule, so I placed it before the membership check.

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -2018,7 +2018,62 @@ describe('deployment app', () => {
...
+          isAdmin,
+          'my reason'
+        );
+
+        await expect(call).rejects.toThrow(
+          ForbiddenErrorCode.CantCancelBundleProduct
+        );
+
+        const untouchedChild =
+          await DeploymentRequestDomain.loadDeploymentRequestBy({
```
carinelebas: [nit] shouldn't we use a test helper here and for untouchedBundle since it's to validate the test result?

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -385,63 +385,44 @@ export const DeploymentApp = {
...
-      });
+    const family =
+      await DeploymentRequestDomain.loadDeploymentRequestWithChildren(
+        deploymentRequest
+      );
 
-    if (!updatedDeploymentRequest) {
-      throw new Error(NotFoundErrorCode.DeploymentRequestNotFound);
-    }
+    await withTransaction(async () => {
```
carinelebas: great idea to add the transaction. However, ideally this transaction should also contain the mail and telemetry sending (which are simply enqueuing events in postgresql) to try to ensure that cancellation and notifications are done in all or nothing mode (but not the call to auth0, which is a sync external call and should not be done in a transaction).

apps/backend/src/modules/deployment/deployment.domain.test.ts
```diff
@@ -254,6 +254,72 @@ describe('deploymentRequestDomain', () => {
     });
   });
 
+  describe('loadDeploymentRequestWithChildren', () => {
+    afterEach(async () => {
+      await DeploymentRequestDomain.deleteDeploymentRequestBy({});
+      await ServiceInstanceDomain.deleteServiceInstanceBy({});
```
carinelebas: [nit] should'nt we use a test helper for this ?

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -56,6 +56,25 @@ export const DeploymentRequestDomain = {
...
+    }
+
+    const children = await DeploymentRequestDomain.loadDeploymentRequestsBy({
+      parent_id: deploymentRequest.id,
+    });
+
+    return [
+      deploymentRequest,
+      ...children.sort((a, b) =>
+        (a.platform_identifier ?? '').localeCompare(b.platform_identifier ?? '')
```
carinelebas, line 78: is the sorting on children really necessary ?

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -385,63 +385,44 @@ export const DeploymentApp = {
...
-            firstName: formatName(requester.first_name ?? ''),
-            platformIdentifier: updatedDeploymentRequest.platform_identifier,
-          },
-        });
-      } else if (!requester) {
-        logApp.warn('Requester not found for trial cancellation mail', {
-          deploymentRequestId: updatedDeploymentRequest.id,
+    for (const previousRequest of family) {
+      const updatedRequest =
+        await DeploymentRequestDomain.loadDeploymentRequestBy({
```
carinelebas: it may be more optimized to reload all the family in one query

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -433,87 +433,37 @@ export const DeploymentApp = {
...
       } catch (error) {
         logApp.error('Error during trial expiration', {
           error,
           deploymentRequestId: trial.id,
         });
+        continue;
+      }
+
+      for (const request of family) {
+        const updatedRequest =
```
carinelebas: it would be more optimized to load all the family in one query

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -435,8 +435,11 @@ export const DeploymentApp = {
     for (const trial of expiredTrials) {
       logApp.info('expiring trial', { deploymentRequestId: trial.id });
 
-      const family =
-        await DeploymentRequestDomain.loadDeploymentRequestWithChildren(trial);
+      const family = (
+        await DeploymentRequestDomain.loadDeploymentRequestWithChildren(trial)
+      ).filter(
+        ({ hub_status }) => hub_status === DeploymentRequestHubStatus.Active
```
carinelebas: why not make this filter directly in the query ? even if it means passing a parameter? it would avoid sending more data than necessary from the DB to the backend.

carinelebas (commented):
nice commit split, it helps with the review 🙏

apps/backend/tests/helper/test.deployment.helper.ts
```diff
@@ -35,6 +35,11 @@ export const TestDeploymentHelper = {
     delete: async (field: DeploymentRequestMutator) => {
       await db<DeploymentRequest>('DeploymentRequest').where(field).del();
     },
+    deleteAllWithServiceInstanceAndSubscription: async () => {
+      await DeploymentRequestDomain.deleteDeploymentRequestBy({});
+      await ServiceInstanceDomain.deleteServiceInstanceBy({});
```
carinelebas: I don't think deleting all serviceInstances and subscriptions when you want to delete a deployment is the best solution. Service instances and subscriptions are not only deployment related. There is no side effect because we re-seed all at each new file, but having a generic helper doing this could be a problem later. 
It thinks we should either clean only ServiceInstance and subscriptions related to deployments, or clean up in the tests only what was inserted.

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -102,9 +102,7 @@ describe('deployment app', () => {
   });
 
   afterEach(async () => {
-    await DeploymentRequestDomain.deleteDeploymentRequestBy({});
-    await ServiceInstanceDomain.deleteServiceInstanceBy({});
-    await TestHelper.subscription.delete({});
+    await TestHelper.deploymentRequest.deleteAllWithServiceInstanceAndSubscription();
     vi.resetAllMocks();
```
carinelebas: This is already done in the general afterEach, there is no need to add it here.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -69,6 +69,9 @@ export const DeploymentRequestDomain = {
       ...(childrenHubStatus ? { hub_status: childrenHubStatus } : {}),
     });
 
+    // Children are sorted by platform_identifier to keep a stable quota lock
```
carinelebas, line 72: 💯

carinelebas (approved):

PR #3127, opened 2026-08-18
feat(user-management): improve pending user approval flow across backend and frontend (#3117)

Kroustille:
# Context
>

This change completes the pending-user moderation flow end-to-end: backend now supports single-user approval with safe/no-op behavior, and frontend has been refactored to use generated GraphQL hooks with clearer, testable pending-user UI logic.

## Changes
1. **Backend**: added acceptPendingUserInOrganization GraphQL mutation, transaction-safe pending lock/read logic, resolver wiring, generated resolver types update, and resolver tests for success + no-op cases.  
2. **Frontend**: migrated pending-user data/mutations to generated GraphQL hooks and query keys, split pending-user UI into dedicated components/hooks, and updated notification/list refresh behavior.  
3. **UX & i18n**: added approve/deny success messages and email-aware confirmation dialogs in en/fr/ja, plus new/updated unit tests for page behavior, dialog handling, and pending-user actions.

1. As an org admin, open Users → Pending users and verify pending requests load.
2. Approve one pending user and confirm: success toast appears, user disappears from pending list, and appears in organization users.
3. Reject one pending user and confirm: success toast appears, user disappears from pending list.
4. Open a direct action URL (with action=approve|deny&user_id=<id>) and verify the correct confirmation dialog opens, then query params are removed.
5. Check notifications: pending-user count updates after approve/reject actions and after subscription invalidation events.

## Related issues

- Related to #3117 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests
- [x] Local manual tests

# Additional information


apps/backend/src/modules/organization-management/user/user-organization/user-organization.app.ts
```diff
@@ -159,6 +159,62 @@ export const UserOrganizationApp = {
...
+      await UserHelper.acceptPendingUserWithCapabilities({
+        user_id: userId,
+        organization_id: organizationId,
+        orgCapabilities: [],
+      });
+
+      return true;
+    });
+
+    if (!pendingUserAccepted) {
```
carinelebas: I wonder why this is needed? should not the code above add the user in the organization ? why do we want to check with an additional query ?

apps/frontend/src/components/admin/user/pending-user/pending-user-list.types.ts
```diff
@@ -0,0 +1,8 @@
+import { UserList_fragment$data } from '@generated/UserList_fragment.graphql';
+
+export type PendingUserAction = 'approve' | 'deny';
```
carinelebas, line 3: question: we usually use enum in the code rather than union type. not a big deal, but I wonder if we should stick to enum, or have both, depending on the cases?

apps/frontend/src/components/admin/user/pending-user/PendingUserConfirmDialog.tsx
```diff
@@ -0,0 +1,55 @@
...
+  pendingUserDialog,
+  onOpenChange,
+  onConfirm,
+}: PendingUserConfirmDialogProps) => {
+  const t = useTranslations();
+  if (!pendingUserDialog) {
+    return null;
+  }
+
+  const confirmModalI18nKey =
```
carinelebas: I'm thinking it could be clearer to have a configuration object with the different values rather than these 3 const with the same if? Wdyt ?
```
const CONFIRM_MODAL_CONFIG: Record<
  PendingUserAction,
  {
    buttonVariant: 'default' | 'destructive';
    icon: JSX.Element;
    i18nKey: string;
  }
> = {
  approve: {
    buttonVariant: 'default',
    icon: <CheckIcon className="h-4 w-4" />,
    i18nKey: 'PendingUserListPage.WarningUserAccept',
  },
  reject: {
    buttonVariant: 'destructive',
    icon: <CloseIcon className="h-4 w-4" />,
    i18nKey: 'PendingUserListPage.WarningUserRejection',
  },
};
```

carinelebas (approved):

apps/backend/src/modules/organization-management/user/user-organization/user-organization.app.ts
Kroustille: The check is here in the case user is not accepted because he does not belong to the requester organization.

The first check with pending user check if he has to be accepted and if so, he's accepted, otherwise it checks if the user is in the requester organization and if no, no success message is displayed to the end user.

Do you think this flow is confusing and needs to be improved ?

apps/frontend/src/components/admin/user/pending-user/pending-user-list.types.ts
Kroustille: I think in this case a union type is a better choice because:

- It's parsed from a raw string (`parsePendingUserAction` in `use-pending-user-dialog.ts:12-14`), likely from a URL param/query string — unions compare directly against string literals with no casting needed.
- Only 2 values, used purely for local state/type-narrowing (dialog state, hook params) — no runtime iteration, no need to pass it around as a namespaced value.
- No GraphQL/backend enum backs it, so there's no codegen consistency argument for using enum.

An enum PendingUserAction `{ Approve = 'approve', Deny = 'deny' }` would only add ceremony (import the enum everywhere, PendingUserAction.Approve vs 'approve') without benefit here. Union of string literals is the right, idiomatic choice — no change needed.

apps/backend/src/modules/organization-management/user/user-organization/user-organization.app.ts
Kroustille: As discussed, check will be moved to `loadUserBy` and throw a user not found error when user does not belong to organization (rejected, deleted or never linked to the organization)

PR #3156, opened 2026-08-20
feat(bundle): bundle quota and eligibility (#3097)

Julienger0:
# Context
>
Voici les quatre descriptions, en anglais simple.

feat(quota): add bundle region quota:
Adds a new limit on how many bundle trials each region can host, alongside the limits that already exist for each product. The new limit starts at the same value as the limit for that region, and trials already running there are counted as taking up space. If more trials are already running than the new limit allows, the number of free spots can start below zero. That is expected and handled.

refactor(quota): key quotas by type instead of platform identifier
Changes the way the code looks up a limit. Until now a limit was always found by product and region, which left no way to point at the new region-wide bundle limit, since that one belongs to no particular product. Limits are now looked up in a way that also says whether it is a bundle limit or a product limit. When several limits have to be updated together, they are always taken in the same order, so two requests arriving at the same time cannot end up waiting on each other forever. Nothing changes for people using the product.

feat(quota): take and release bundle quota alongside product quota:
Keeps both limits in step. Asking for a single product trial now takes a spot from that product's limit and from the region's bundle limit, because that trial really does occupy a place in the region. Asking for a bundle takes only a bundle spot. The products inside a bundle take nothing on their own, since the bundle already accounts for them. The same rule gives the spots back when a trial is cancelled or expires, so a request can never take from one limit and give back to another. Only the limit that truly belongs to the request can turn it down: if the product limit is full, the request waits in the queue. The bundle spot is always taken, even when that pushes the count below zero.

feat(bundle): restrict bundle trial eligibility per organization:

Sets the rules for who can ask for a bundle trial. An organization can have only one, and cannot ask for one while it still has a single-product trial running. A product trial that was cancelled or has expired no longer stands in the way. The rules for single-product trials are unchanged. The check is now repeated at the moment the request is saved, so two people from the same organization asking at the same time can no longer both succeed.

## Related issues

- Related to #3097

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/deployment/quota/deployment.quota.domain.ts
```diff
@@ -18,22 +46,45 @@ export const isQuotaManagedPlatform = (
   platformIdentifier !== null &&
   platformIdentifier !== PlatformIdentifier.Xtmone;
 
+export const isQuotaManagedKey = (key: QuotaKey): boolean =>
+  key.type === DeploymentRequestDeploymentType.Bundle ||
+  isQuotaManagedPlatform(key.platformIdentifier);
```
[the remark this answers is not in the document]
Julienger0: false positive

apps/backend/src/migrations/20260819100000_add_bundle_deployment_quota.js
```diff
@@ -0,0 +1,96 @@
...
+      indexName: 'deployment_request_quota_region_bundle_unique',
+      predicate: knex.whereRaw("type = 'bundle'"),
+    });
+  });
+
+  const openctiQuotas = await knex('DeploymentRequestQuota')
+    .where('platform_identifier', '=', 'opencti')
+    .select('region', 'capacity');
+
+  const occupancyRows = await knex('DeploymentRequest')
```
carinelebas: I think i t would be easier to compute the occupancy from the quota table rather than from the deploymentRequests. The logic would be simpler and less error prone.

apps/backend/src/modules/deployment/quota/deployment.quota.domain.ts
```diff
@@ -9,7 +10,34 @@ import DeploymentRequestQuota, {
 } from '../../../model/kanel/public/DeploymentRequestQuota';
 import { ErrorCode } from '../../../utils/error/error.code';
 
-export const isQuotaManagedPlatform = (
+export type QuotaKey =
```
carinelebas: maybe this could be simplifed to `type QuotaKey = { type; platformIdentifier: PlatformIdentifier | null; region };` ? 
it would allow removing `quotaKeyPlatformIdentifier` and simplify `compareQuotaKeys` and `scopeToQuotaKey` ? 

we should keep in mind that product quota will disappear when we can only create bundle requests.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -110,6 +113,10 @@ export const DeploymentApp = {
 
     try {
       if (validatedProducts.type === DeploymentRequestDeploymentType.Bundle) {
+        await DeploymentHelper.assertNoExistingBundle(
```
carinelebas: this check should actually be done (or done again) within the transaction to avoid 2 concurrent creations to be successful.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -23,6 +22,18 @@ import {
   ServiceGroupDomain,
   ServiceGroupName,
 } from './group/service-group.domain';
+import { QuotaKey } from './quota/deployment.quota.domain';
+
+const scopeToQuotaKey =
```
carinelebas: since the quotaKey also contains the region, I would expect this function to also set the region? wdyt?

apps/backend/src/modules/deployment/quota/deployment.quota.app.ts
```diff
@@ -0,0 +1,91 @@
...
+  QuotaRequestShape,
+} from './deployment.quota.domain';
+
+const QUOTA_HOLDING_HUB_STATUSES = [
+  DeploymentRequestHubStatus.Active,
+  DeploymentRequestHubStatus.Pending,
+  DeploymentRequestHubStatus.Provisioning,
+];
+
+export const DeploymentQuotaApp = {
```
carinelebas, line 104: all of this seems a bit complex to manage the quota sync, knowing that we'll remove the product quotas once the FF is activated?

apps/backend/src/modules/deployment/deployment.helper.ts
```diff
@@ -138,14 +148,36 @@ export const DeploymentHelper = {
...
+        throw new Error(AlreadyExistsErrorCode.FreeTrialAlreadyExists);
+      }
+
+      const ongoingTrials =
+        await DeploymentRequestDomain.loadOngoingStandaloneTrialsForOrganization(
+          organizationId
+        );
+      if (ongoingTrials.length > 0) {
+        throw new Error(AlreadyExistsErrorCode.FreeTrialAlreadyExists);
+      }
```
carinelebas: I'm not sure this is needed since we'll implement a different behavior later?

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -308,6 +318,29 @@ export const DeploymentRequestDomain = {
...
   },
 
+  lockOrganizationDeploymentRequests: async (
+    organizationId: OrganizationId
+  ): Promise<void> => {
+    const transaction = databaseContext.getTransaction();
+    if (!transaction) {
+      throw new Error(UnknownErrorCode.UnknownError);
+    }
+    await transaction.raw('SELECT pg_advisory_xact_lock(hashtext(?))', [
```
carinelebas: if we later need to set another advisory lock on the organizationId, there may be a collision. 
It would be better to scope the lock with something like `pg_advisory_xact_lock(hashtext('deployment_request'),hashtext(?)) `

Ideally, we could even add this to the transaction framework in database context, to ensure a namespace is always provided, maybe something like:
```
withAdvisoryLock: async <T>(
  namespace: string,
  key: string,
  callback: () => Promise<T>
): Promise<T> =>
  withTransaction(async () => {
    const transaction = getTransaction()!; // always set, we're inside withTransaction
    await transaction.raw(
      'SELECT pg_advisory_xact_lock(hashtext(?), hashtext(?))',
      [namespace, key]
    );
    return callback();
  }),
```
wdyt?

apps/backend/src/modules/deployment/deployment.domain.ts
Julienger0: the region is set inside the function now

apps/backend/src/modules/deployment/deployment.domain.ts
Julienger0: Agreed on both counts, and I took your second suggestion, withAdvisoryLock(namespace, key, callback) is now in the database context, so a namespace is always required and the deployment lock uses 'deployment_request'

apps/backend/src/modules/deployment/quota/deployment.quota.app.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (promotedRequest) {
+    return promotedRequest;
+  }
+
+  await DeploymentQuotaDomain.freePlace(key);
+  return undefined;
+};
+
+const releaseStandaloneTrialQuota = async (
```
carinelebas: This function allows to promote one bundle + one standalone trial. However, we release only one slot. 
Even if the 'bundle' quota is not blocking when requesting a trial, releasing trial space can happen after the feature is activated. And in this case, the 'bundle' quota should be the one counting, it may not be respected.
-> This is an edge case, but I think we should only promote one product, since we release only one space, either bundle or standalone. I think we should first promote bundle if any, otherwise standalone. I think we won't have queued standalone when activating the bundles. 

Even this is anyway a dangerous logic, because if the bundle quota is negative, and we keep promoting bundles though trials spaces, it will never go back positive 🤔

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -2700,6 +2747,472 @@ describe('deployment app', () => {
...
+    });
+
+    const loadAvailability = async (
+      filter: Parameters<typeof TestHelper.deploymentRequestQuota.load>[0]
+    ) => {
+      const quota = await TestHelper.deploymentRequestQuota.load(filter);
+      return quota!.availability;
+    };
+
+    const setAvailability = async (
```
carinelebas: question: are these wrappers really useful?

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -2700,6 +2747,472 @@ describe('deployment app', () => {
     });
   });
 
+  describe('bundle quota', () => {
```
carinelebas: I think these tests should not be in a dedicated section, but on the main functions tests that they are calling, for example  quota dedicated section in `createDeploymentRequest` tests for tests calling this function. It would avoid risks of test duplication

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -2700,6 +2747,472 @@ describe('deployment app', () => {
...
+          parent_id: bundle.id as DeploymentRequestId,
+        });
+        expect(
+          children.map((child) => child.platform_identifier).sort()
+        ).toEqual([PlatformIdentifier.Opencti, PlatformIdentifier.Xtmone]);
+      });
+    });
+
+    describe('release', () => {
+      it('should refuse to cancel a bundle child on its own', async () => {
```
carinelebas: This test a duplication of another one.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -453,30 +455,21 @@ export const DeploymentApp = {
...
-        user.id
-      );
+    if (promotedRequests.length === 0) {
       return;
     }
 
-    await DeploymentQuotaDomain.freePlace(key);
+    const user = requestContext.requireUser();
+    for (const promotedRequest of promotedRequests) {
+      await sendUpdateDeploymentTelemetryEvent(promotedRequest, user.id);
```
carinelebas, line 472: this should be done in the same transaction where request is promoted to ensure the all or nothing behavior.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -269,6 +269,17 @@ export const DeploymentRequestDomain = {
       })
       .where('id', '=', request.id)
       .returning('*');
+
```
carinelebas, line 283: no test for this?

apps/backend/src/context/database.context.ts
```diff
@@ -27,6 +28,23 @@ export const databaseContext = {
     });
   },
 
+  withAdvisoryLock: async <T>(
```
carinelebas, line 31: nice improvement, but no tests ?

apps/backend/src/modules/deployment/deployment.app.test.ts
Julienger0: setAvailability wasn't but I kept loadAvailability is useful in my opinion

apps/backend/src/modules/deployment/quota/deployment.quota.app.ts
Julienger0: yes fixed and I added that test  :"should promote the queued bundle rather than the queued trial, and give the product place back"

apps/backend/src/modules/deployment/deployment.app.test.ts
Julienger0: agreed

apps/backend/src/modules/deployment/deployment.app.ts
Julienger0: That block runs inside the transaction but i made somes changes, requestContext.requireUser() and  the loop is now a single if since the release promotes at most one request

apps/backend/src/modules/deployment/deployment.domain.ts
Julienger0: added : "hould cascade to the queued children when the promoted request is a bundle"

apps/backend/src/context/database.context.ts
Julienger0: tests added

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -2212,108 +2406,316 @@ describe('deployment app', () => {
       }
     );
 
-    describe('bundle cancellation', () => {
-      let bundle: DeploymentRequest;
-      let childOpencti: DeploymentRequest;
-      let childXtmone: DeploymentRequest;
-
+    describe('quota', () => {
```
carinelebas, line 2533: should it not be in the `bundle cancellation` describe?

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -482,20 +482,16 @@ export const DeploymentApp = {
...
+    const promotedRequest = await DeploymentQuotaApp.releaseQuotaForRequest(
       request,
       previousHubStatus
     );
 
-    if (promotedRequests.length === 0) {
-      return;
-    }
-
-    const user = requestContext.requireUser();
```
carinelebas: question: why don't you use the user from the context anymore?

apps/backend/src/context/database.context.test.ts
```diff
@@ -174,4 +174,148 @@ describe('databaseContext Tests', () => {
       expect(await countItems()).toBe(0);
     });
   });
+
+  describe('advisory lock', () => {
```
carinelebas, line 178: 💯

apps/backend/src/modules/deployment/deployment.app.test.ts
Julienger0: Only one test in that block is a bundle cancellation
 I moved 'should give back only the bundle place when a bundle with an active child is cancelled' into a nested quota describe under bundle cancellation
 The other five cancel a standalone trial I put them under cancelDeploymentRequest
 Should I move the others as well?

apps/backend/src/modules/deployment/deployment.app.ts
Julienger0: Because of the trial expiry cron: there the context user is the CRONS user, while the expiration is already reported with SYSTEM_USER_UUID, passing the id explicitly keeps both consistent

apps/backend/src/modules/deployment/deployment.app.test.ts
carinelebas: I think it's good ! This file is getting too big 😄 thanks!

apps/backend/src/modules/deployment/quota/deployment.quota.app.ts
```diff
@@ -59,10 +101,27 @@ const releaseStandaloneTrialQuota = async (
 };
 
 export const DeploymentQuotaApp = {
-  takeBundleQuota: (
-    region: DeploymentRequestPlatformRegion
-  ): Promise<{ isPlaceAvailable: boolean }> =>
-    DeploymentQuotaDomain.reservePlace(bundleQuotaKey(region)),
+  takeBundleQuota: async (
```
carinelebas: `updateDeploymentQuotaCapacity` doesn't call this function, so this logic is not applied when updating quotas. 

I think we should have 2 entrypoint function that manage the different cases one to take the quotas, and one to release them:
- these functions could manage always the same way taking/releasing the quotas, depending if it's a bundle, a standalone trial or bundle trial
- all caller function should exclusively call one of these 2 functions
- reservePlace/freePlace etc should not be use anymore outside if the quota domain/app. 

It would make the logic clearer, because currently each case is going through different flow, it's very difficult to follow.
it could also prevent some cases to be forgotten.

Also, this function and other new functions have no test.

apps/backend/src/modules/deployment/deployment.app.ts
carinelebas: I'm not sure to understand what consistency you mean. Getting the use from the context is to know automatically which user triggered the action. Where do we need consistency?

Julienger0: nevermind I was completely wrong on that, misunderstood how it was working
i put back the user context

apps/backend/src/modules/deployment/quota/deployment.quota.app.test.ts
```diff
@@ -0,0 +1,489 @@
...
+      );
+
+      expect(promoted).toBeUndefined();
+      expect(
+        await loadAvailability(productFilter(PlatformIdentifier.Opencti))
+      ).toBe(6);
+      expect(await loadAvailability(bundleFilter)).toBe(5);
+    });
+
+    it('should promote a queued trial and take a bundle place when a bundle child is released', async () => {
```
carinelebas: question : I wonder if this is something we want to do? 
when we release a bundle, we already try to promote a bundle. With this logic of child promotion too, we may additionally promote a standalone trial 🤔

apps/backend/src/modules/deployment/deployment.domain.test.ts
```diff
@@ -1185,7 +1189,16 @@ describe('deploymentRequestDomain', () => {
     });
   });
 
-  describe('setFirstQueuedRequestAsPending', () => {
+  describe('loadFirstQueuedRequest and setRequestAsPending', () => {
```
carinelebas: [nit] since the functions are now distinct, they should be tested independently. Or do you think we will merge them again when cleaning up product quotas?

carinelebas (approved):
great job! logic is easier to understand, so easier to maintain and less bug prone! Well tested too! 🎉

apps/backend/src/modules/deployment/quota/deployment.quota.app.test.ts
Julienger0: Agreed, I made the change

apps/backend/src/modules/deployment/deployment.domain.test.ts
```diff
@@ -1189,7 +1189,98 @@ describe('deploymentRequestDomain', () => {
...
+        trialQuotaKey(PlatformIdentifier.Opencti, openctiRequest!.region)
+      );
+
+      expect(request?.id).toBe(openctiRequest!.id);
+    });
+  });
+
+  describe('setRequestAsPending', () => {
     const promoteFirstQueuedRequest = async (key: QuotaKey) => {
       const request = await DeploymentRequestDomain.loadFirstQueuedRequest(key);
```
carinelebas: why do we still call loadFirstQueuedRequest in tests for setRequestAsPending ?

carinelebas (approved):

PR #3164, opened 2026-08-21
feat(frontend): add XTM Platform trial pages and request form (#3099)

Antoine-Quintin:
# Context
>

public page and private form for admin user without standalone trial 
[Kooha-2026-08-21-16-21-57.webm](https://github.com/user-attachments/assets/c0c7ab2a-3c94-4065-8511-975e836b8bfe)

**public page**

image

**admin user, orga without trial**

image

**admin user, orga with trial**

image

**admin user in its personnal space**

image

**non admin user**

image

Adds the XTM Platform trial pages and request form (chunk 8a of the XTM Platform Trial feature).

**Pages** — a public page at `/cybersecurity-solutions/xtm-platform-trial` and a private one at `/app/service/xtm-platform-trial`, both built around a shared pitch and a right-hand panel that adapts to the visitor:

- not logged in → sign-in panel
- logged in without `ADMINISTRATE_ORGANIZATION` or `MANAGE_PLATFORM_REGISTRATION` → explanatory message
- allowed → the bundle request form

**Form** — product checkboxes (XTM One always included and not selectable), one use case select per selected product, plus region, job title, activity sector and the MSSA acknowledgement. Submitting creates a bundle deployment request.

**`use_case` → `use_cases_by_product`** — a bundle carries one use case per product, which the previous single `use_case` input could not express. Following Carine's review, `use_case` is removed from `CreateDeploymentRequestInput` entirely and the list becomes the only source, for standalone trials too. The database column is unchanged: only the way the client expresses the use case evolves. The bundle parent keeps a null use case since its children carry theirs.

**Cancelling ongoing standalone trials** — when an organization requests a bundle while it still has standalone trials running, those trials are cancelled **once the bundle reaches the Active state**, not at request time: the pending window is short and the bundle may never activate. The form warns the user before submission, and no cancellation email is sent since they receive the bundle activation email instead. Only standalone trials of the same organization are affected — bundle children are excluded, and the cancellation only runs on the transition to Active, so updating an already active bundle cancels nothing again.

## How to test

**Request form**
1. Log in as an organization admin and open `/app/service/xtm-platform-trial`
2. All three products are checked by default; unchecking OpenCTI or OpenAEV removes its use case select, rechecking brings it back empty
3. The submit button is disabled when neither OpenCTI nor OpenAEV is selected
4. Submit and check in the database: the bundle parent has `use_case: null`, each child carries the use case of its product, the XTM One child has none

**Panel states**
5. Log out and open `/cybersecurity-solutions/xtm-platform-trial` → sign-in panel
6. Log in as a non-admin member → "contact your organization admin" message

**Standalone cancellation**
7. Create a standalone OpenCTI trial, then open the bundle page → the products block shows a red warning instead of the usual one
8. Request the bundle, then set the bundle to Active → the standalone is cancelled with the automatic reason

## Related issues

- Related to #3099

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/src/components/ui/TranslatableEnumSelectField.tsx
```diff
@@ -9,25 +9,27 @@ import {
...
+  field: ControllerRenderProps<T, FieldPath<T>>;
   label: string;
   placeholder: string;
   values: string[];
   translationNamespace: string;
   className?: string;
 }
 
-export const TranslatableEnumSelectField = ({
+export const TranslatableEnumSelectField = <T extends FieldValues>({
```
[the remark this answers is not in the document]
Antoine-Quintin: The extends clause is enough for the TSX parser to disambiguate from JSX — the trailing comma is only needed for a bare < T >. yarn check-ts passes as is.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -107,6 +107,21 @@ export const DeploymentRequestDomain = {
...
   },
 
+  loadOngoingStandaloneTrialsForOrganization: async (
+    organizationId: OrganizationId
+  ): Promise<DeploymentRequest[]> => {
+    return db<DeploymentRequest[]>('DeploymentRequest')
+      .where('organization_requester_id', '=', organizationId)
+      .where('type', '=', DeploymentRequestDeploymentType.Trial)
+      .whereNull('parent_id')
+      .whereNotIn('hub_status', [
```
hervyt: **question**: is it dangerous if we ever add a status? What about a "whereIn" rather ?

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -107,6 +107,21 @@ export const DeploymentRequestDomain = {
...
+        DeploymentRequestHubStatus.Cancelled,
+        DeploymentRequestHubStatus.Expired,
+        DeploymentRequestHubStatus.Failed,
+      ])
+      .select('*');
```
hervyt, line 142: **thought**: Do you really need all columns? We are in times where we have few perf issues (minor, yes, but for the future, when we will have laaaarge table (one day...))

apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+export async function generateMetadata({
+  params,
+}: {
+  params: Promise<{ locale: PublicLocale }>;
+}): Promise<Metadata> {
+  const { locale } = await params;
+  const baseUrl = await getBaseUrl();
+  const t = await getTranslations({
+    locale,
+    namespace: 'Service.Trials.XtmPlatform.Page',
```
Comment (hervyt) [inline apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx]: **nitpick**: I thought the team decision was to use getTranslation from root and everytime in the "t" add the full route. (+: if we move on tolgee, it will be easier to migrate! ;) )

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialLimitations.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { InfoIcon } from '@filigran/icon';
+import { useTranslations } from 'next-intl';
+
+export const XtmPlatformTrialLimitations = () => {
+  const t = useTranslations('Service.Trials.XtmPlatform.Page.Limitations');
+
+  return (
+    <div className="rounded p-6 bg-elevation-background-layer-1">
+      <h3 className="flex items-center gap-s text-primary mb-s font-bold">
```
hervyt: **A11Y**: This is an h3 but the only title I see above is an h1, so in term of A11Y, we miss the h2! 
+ With design system, we shouldn't rewrite <h> styles.... Can't wait the end of september to have propers styles!

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPitch.tsx
```diff
@@ -0,0 +1,50 @@
...
+];
+
+const PRODUCT_TRANSLATION_KEYS: Record<PlatformIdentifier, string> = {
+  [PlatformIdentifier.Opencti]: 'opencti',
+  [PlatformIdentifier.Openaev]: 'openaev',
+  [PlatformIdentifier.Xtmone]: 'xtmone',
+};
+
+export const XtmPlatformTrialPitch = () => {
+  const t = useTranslations('Service.Trials.XtmPlatform.Page');
```
hervyt: **thought**: Same here (and other places) for translation ;)

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPitch.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useTranslations } from 'next-intl';
+
+const BUNDLE_PRODUCTS = [
+  PlatformIdentifier.Opencti,
+  PlatformIdentifier.Openaev,
+  PlatformIdentifier.Xtmone,
+];
+
+const PRODUCT_TRANSLATION_KEYS: Record<PlatformIdentifier, string> = {
+  [PlatformIdentifier.Opencti]: 'opencti',
```
hervyt: **question**: Why don't you use PlatformIdentifier directly ? The PlatformIdentifier is directly opencti/openaev/xtmone, no? 🤔

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPitch.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <h2 className="heading-xl">{t('PitchTitle')}</h2>
+        <p className="content-body-base">{t('PitchDescription')}</p>
+      </div>
+
+      {BUNDLE_PRODUCTS.map((platformIdentifier) => {
+        const { name, Icon } = PlatformMetadataMapping[platformIdentifier];
+        const productKey = PRODUCT_TRANSLATION_KEYS[platformIdentifier];
+
+        return (
+          <article
```
hervyt: 💯

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPage.tsx
```diff
@@ -5,25 +5,34 @@ import { ReactNode } from 'react';
...
+        <p className="heading-sm bg-clip-text text-transparent bg-gradient-focus">
+          {t('Overline')}
+        </p>
         <h1 className="heading-2xl">{t('Title')}</h1>
       </header>
 
-      <div className="grid grid-cols-1 lg:grid-cols-2 gap-xl px-m pb-xxl">
-        <div className="flex flex-col gap-xl">
+      <div className="flex flex-col gap-l lg:flex-row">
+        {' '}
```
hervyt: **nitpick**: Do you really need this?

apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx
```diff
@@ -71,7 +72,7 @@ const Page = async ({
         dangerouslySetInnerHTML={{ __html: stringifyJsonLd(jsonLd) }}
       />
       <BreadcrumbNav value={breadcrumbs} />
-      <XtmPlatformTrialPage panel={null} />
+      <XtmPlatformTrialPage panel={<PublicXtmPlatformTrialPanel />} />{' '}
```
Comment (hervyt) [inline apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx]: **nitpick**: Do you really need the space?

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialMessagePanel.tsx
```diff
@@ -0,0 +1,24 @@
...
+  actions?: ReactNode;
+}
+
+export const XtmPlatformTrialMessagePanel = ({
+  title,
+  description,
+  actions,
+}: XtmPlatformTrialMessagePanelProps) => (
+  <div className="flex h-full flex-col gap-l rounded p-l bg-elevation-background-layer-2">
+    {' '}
```
hervyt: **nitpick**: Again a space? 👅

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -599,30 +599,45 @@ type ValidatedDeploymentRequestProducts =
       platformIdentifier: PlatformIdentifier;
     };
 
+const validateUseCasesByProduct = (
```
hervyt: **thoughtùù: Maybe we should do it in a helper, not app.ts?

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/PrivateXtmPlatformTrialPanel.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return null;
+  }
+
+  const state = deriveXtmPlatformTrialPanelState({
+    isAllowed: canRequestTrial,
+    ongoingStandaloneTrials:
+      data?.platformTrialStatus?.ongoingStandaloneTrials ?? [],
+  });
+
+  if (state === 'not-allowed') {
```
hervyt: **question**: Do you have this value elsewhere? If not, we could use an enum not to have magicStrings

apps/frontend/src/hooks/use-can-request-platform-trial.ts
```diff
@@ -0,0 +1,24 @@
+import useGranted from '@/hooks/use-granted';
+import { useAdminByPass } from '@/hooks/use-portal-capability';
+import { OrganizationCapability } from '@graphql/generated';
+
+/**
+ * Mirrors the server guard on createDeploymentRequest:
+ * ADMINISTRATE_ORGANIZATION or MANAGE_PLATFORM_REGISTRATION,
+ * with a bypass for platform administrators.
+ */
+export const useCanRequestPlatformTrial = (): boolean => {
```
hervyt: You already have an useOrganizationCapabilities that already exists. Can't we reuse it instead of hyper specialized hooks?

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+        </span>
+        <span className="txt-default-bold">{me?.email}</span>
+      </div>
+
+      <Form {...form}>
+        <form
+          className="flex w-full flex-col gap-l"
+          onSubmit={form.handleSubmit(handleSubmit)}>
+          <div className="flex flex-col gap-s rounded bg-hover p-l">
+            <h3 className="txt-title-xs">
```
hervyt, line 169: Same here, an h3 without an h1 or h2 above 😉

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+  return (
+    <div className="flex w-full flex-col gap-xl rounded bg-page-background p-xl">
+      <div className="flex items-center gap-l">
+        <span className="txt-default text-muted-foreground">
+          {t('Service.Trials.Form.AssociatedEmail')}
+        </span>
+        <span className="txt-default-bold">{me?.email}</span>
+      </div>
+
+      <Form {...form}>
```
hervyt, line 164: **question**: You should use AutoForm 📦

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/PrivateXtmPlatformTrialPanel.tsx
```diff
@@ -83,14 +84,10 @@ export const PrivateXtmPlatformTrialPanel = () => {
...
-        <XtmPlatformTrialForm handleSubmit={handleSubmit} />
-      </>
-    );
-  }
-
-  return <XtmPlatformTrialForm handleSubmit={handleSubmit} />;
+  return (
+    <XtmPlatformTrialForm
+      handleSubmit={handleSubmit}
+      hasOngoingStandaloneTrials={state === 'request-with-ongoing-trials'}
```
hervyt: Same here, can use use an enum not to have magic strings ? :)

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -1278,6 +1280,24 @@ const recomputeBundleDates = async (bundleId: DeploymentRequestId) => {
   );
 };
 
+const cancelOngoingStandaloneTrialsForBundle = async (
```
hervyt: This shoulod be in a helper, not in app.ts

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -1553,6 +1553,149 @@ describe('deployment app', () => {
         });
         expect(active.hub_status).toBe(DeploymentRequestHubStatus.Active);
       });
+      describe('standalone trials cancellation', () => {
+        beforeEach(() => {
+          vi.spyOn(DeploymentQuotaDomain, 'freePlace').mockResolvedValue();
+        });
+
+        it('should cancel ongoing standalone trials of the organization when the bundle becomes active', async () => {
```
hervyt: **nitpick of nitpick**: It can be easier to read/structure with the GIVEN//WHEN//THEN pattern

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/xtm-platform-trial-panel.utils.test.ts
```diff
@@ -7,36 +7,56 @@ import { describe, expect, it } from 'vitest';
...
         'the user is allowed and the organization has one ongoing trial',
+      isPersonalSpace: false,
       isAllowed: true,
       ongoingStandaloneTrials: [PlatformIdentifier.Opencti],
       expected: 'request-with-ongoing-trials',
     },
     {
       description:
         'the user is allowed and the organization has two ongoing trials',
+      isPersonalSpace: false,
```
hervyt, line 70: We can add the test with personal_space=true

apps/backend/src/modules/deployment/deployment.domain.ts
Antoine-Quintin: with whereNotIn any status added later would silently count as ongoing and get cancelled, which is the dangerous default. Switching to whereIn on the non-terminal statuses (queued, pending, provisioning, active) so a new status is ignored until someone decides otherwise.

apps/backend/src/modules/deployment/deployment.domain.ts
Antoine-Quintin: The full row is needed here: this feeds applyCancellationToDeploymentRequest, which reads actual_state, region, platform_identifier and service_instance_id to compute the quota release and the target state. The other caller (loadPlatformTrialStatus) only maps to platform_identifier, so I could split it into two domain functions. The query is scoped to a single organization's ongoing trials, so the volume stays small normally.

apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx
Comment (Antoine-Quintin) [reply apps/frontend/app/(public)/[locale]/cybersecurity-solutions/xtm-platform-trial/page.tsx]: Done

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPitch.tsx
Antoine-Quintin: done

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialLimitations.tsx
carinelebas: it should be heading-xs

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialLimitations.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <InfoIcon className="size-4" />
+        {t('Title')}
+      </h3>
+      <p className="text-sm mb-s">{t('Intro')}</p>
+      <ul className="text-sm mb-l list-disc pl-l">
+        <li>{t('OpenCTI')}</li>
+        <li>{t('OpenAEV')}</li>
+      </ul>
+      <p className="text-sm">{t('Ingestion')}</p>
+      <p className="text-sm">{t('NoSLA')}</p>
```
carinelebas: we should use the typo from the mockup

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialPitch.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { PlatformIdentifier } from '@graphql/generated';
+import { useTranslations } from 'next-intl';
+
+const BUNDLE_PRODUCTS = [
+  PlatformIdentifier.Opencti,
+  PlatformIdentifier.Openaev,
+  PlatformIdentifier.Xtmone,
+];
+
+const PRODUCT_TRANSLATION_KEYS: Record<PlatformIdentifier, string> = {
```
carinelebas: why not use directly the product identifier in the translation file and avoid this mapping ?

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -599,30 +599,45 @@ type ValidatedDeploymentRequestProducts =
...
+      (entry) =>
+        entry.platform_identifier !== PlatformIdentifier.Xtmone &&
+        uniqueProducts.includes(entry.platform_identifier)
+    ) ?? true;
+  if (!everyEntryTargetsAValidProduct) {
+    throw new Error(BadRequestErrorCode.MissingUseCaseForProduct);
+  }
+  if (!everyProductHasUseCase) {
+    throw new Error(BadRequestErrorCode.MissingUseCaseForProduct);
+  }
```
carinelebas: maybe we could have only one if ?

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+        <form
+          className="flex w-full flex-col gap-l"
+          onSubmit={form.handleSubmit(handleSubmit)}>
+          <div className="flex flex-col gap-s rounded bg-hover p-l">
+            <h3 className="txt-title-xs">
+              {t('Service.Trials.XtmPlatform.Page.Form.ProductsTitle')}
+            </h3>
+
+            <div className="flex items-start gap-xs rounded border border-solid border-orange p-s text-sm ">
```
carinelebas: we should not use text-sm

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+              <div className="flex items-center gap-s">
+                <Checkbox
+                  id="product-xtmone"
+                  checked
+                  disabled
+                />
+                <label
+                  htmlFor="product-xtmone"
+                  className="txt-default text-muted-foreground">
```
carinelebas: it should not be txt-default, but content-base

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
carinelebas, line 1: we should use the right typo in this form

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -1553,6 +1553,149 @@ describe('deployment app', () => {
...
+              }
+            );
+
+          await DeploymentApp.updateDeploymentRequest({
+            id: bundle.id as DeploymentRequestId,
+            actual_state: DeploymentRequestPlatformState.Active,
+          });
+
+          const cancelled =
+            await DeploymentRequestDomain.loadDeploymentRequestBy({
```
carinelebas: [nit] we should use a test helper here

apps/backend/src/modules/deployment/deployment.app.test.ts
carinelebas: I don't think we should not cancel the standalone trials when the bundle becomes active, but when we request the bundle. Otherwise, the organization will take 2 quota spaces until the bundle becomes active, and we want to let space for others.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -79,6 +79,8 @@ import { DeploymentHelper } from './deployment.helper';
 import { DeploymentQuotaDomain } from './quota/deployment.quota.domain';
 
 export const XTM_PLATFORM_BUNDLE_SERVICE_INSTANCE_NAME = 'XTM Platform Bundle';
+export const BUNDLE_ACTIVATION_CANCELLATION_REASON =
```
carinelebas: what is the telemetry event sent in this case? we cannot simply add a cancellation reason, it would have impact on marketing side. If we want to set something automatically, it should fit within the current cancellation reasons (maybe in the 'other' part), or be discussed with data and marketing.

carinelebas (commented):
A few additional comments: 
- background color for limitations is not the right one.
- placeholder in the form + MSSA are not the right ones
- we don't display that MSSA ticking is mandatory
- there must be some margin main text next ("Explore the full XTM Platform experience") and the menu
- in the warning text on the form 'Once your trial starts' should be on a new line

apps/frontend/graphql/trial/trial.mutation.graphql
carinelebas, line 1: I think this file should be named trial.mutation.graphql. we don't want a graphql file per query or mutation.

apps/frontend/graphql/trial/trial.query.graphql
carinelebas, line 1: we should also probably rename this file trial.query.graphql

apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialForm.tsx
Kroustille: For this one, there is already h2 in the same page so I'll keep the h3

apps/backend/src/modules/deployment/deployment.app.ts
Kroustille: Since it's a temporary cancellation reason (until there's no standalone trials), I will put it as other cancellation reason 😉

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -547,6 +570,142 @@ describe('deployment app', () => {
...
+          // Then the bundle's own children are untouched
+          expect(children).not.toHaveLength(0);
+          children.forEach((child) => {
+            expect(child.hub_status).not.toBe(
+              DeploymentRequestHubStatus.Cancelled
+            );
+          });
+        });
+
+        it('should not retroactively cancel standalone trials created after the bundle was requested', async () => {
```
carinelebas: question : this test is a bit strange, no? how could it happen ? On top of that, you should not be allowed to create a standalone trial if you have a bundle trial in your orga 🤔

apps/backend/src/modules/deployment/deployment.app.test.ts
```diff
@@ -243,20 +244,24 @@ describe('deployment app', () => {
         ],
       });
 
+      // Then
```
carinelebas: These comments are added every where.. is this change really needed? Are these comments really mandatory in each test? if so we should all align on that, to avoid adding them everywhere. If not, we should not add them on 'untouched' tests? wdyt?

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -416,7 +415,9 @@ export const DeploymentApp = {
     });
 
     for (const previousRequest of family) {
-      await deleteDeploymentRequestAudience(previousRequest);
+      await DeploymentRequestDomain.deleteDeploymentRequestAudience(
```
carinelebas: there is no need to delete audience for xtmone or openaev, they will never be created with an audience.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -15,14 +16,26 @@ import DeploymentRequest, {
...
-import { ErrorCode, UnknownErrorCode } from '../../utils/error/error.code';
+import { getErrorNumberProperty } from '../../utils/error/error-guard.util';
+import {
+  ErrorCode,
+  NotFoundErrorCode,
+  UnknownErrorCode,
+} from '../../utils/error/error.code';
 import { prefixObjectKeys } from '../../utils/utils';
+import { ServiceInstanceDomain } from '../service/instance/service-instance.domain';
+import {
```
carinelebas: [nit] I find it a bit weird that deployment.domain imports from deployment.app, no? could it lead to circular dependencies?

apps/backend/src/modules/deployment/deployment.app.test.ts
Kroustille: Yes I agree it's a strange test and case should not happen. I'm removing it 😉

apps/backend/src/modules/deployment/deployment.domain.ts
Kroustille: Yes I agree, I also found it weird but it was the most straightforward way to move functions across files but it can show us that we have a responsibility issue within both files. I'm checking if I can make it clear without much duplications 😉

apps/backend/src/modules/deployment/deployment.app.ts
Kroustille: I added a check into `deleteDeploymentRequestAudience` to only delete audience for OpenCTI deployment requests.

apps/backend/src/modules/deployment/deployment.domain.ts
```diff
@@ -541,6 +540,10 @@ export const DeploymentRequestDomain = {
   deleteDeploymentRequestAudience: async (
     deploymentRequest: DeploymentRequest
   ): Promise<void> => {
+    if (deploymentRequest.platform_identifier !== PlatformIdentifier.Opencti) {
```
carinelebas: I'm not sure we should be so restrictive here at that level? 
In the case of a bundle, we are not creating audiences for openaev/xtmone. 
However, for current trials, we were creating audiences until 2 weeks ago (ref [this PR](https://github.com/XTM-Hub/xtm-hub/pull/3019)), so we still have openaev trials with audience? Is this function also called for standalone trials ?

hervyt (approved):

carinelebas (approved):

PR #3185, opened 2026-08-24
chore(ai): add cloud agent setup, path-scoped instructions and a review skill (#3185)

jpkha:
# Context

This improves how coding agents (Copilot and others) work in this repository. The problems, in order of how they surfaced:

1. **The Copilot coding agent could not validate anything.** There was no `copilot-setup-steps.yml`, so the cloud agent booted without `corepack enable` or `yarn install` and could never run `check-ts`, `lint` or `test` before opening a PR.
2. **`copilot-instructions.md` was 26 KB / 561 lines and loaded on every single request**, so backend rules were burning context while editing frontend files, and vice versa.
3. **Several facts in it had drifted and were actively misleading** (a `.rules` file and a `generate:enum` script that do not exist, wrong Node/Yarn/Next versions, a stale CI job graph).
4. **The instructions kept re-drifting as I edited them.** Rules got updated in one file but not in every place that repeated them (see "Contradictions found and fixed" below), and once path-scoped instructions and custom agents both existed, they started overlapping in what they told an agent to do.

## Approach

**`.github/workflows/copilot-setup-steps.yml`** preinstalls the toolchain for the cloud agent and runs the codegen the repo needs before type-checking works at all: `generate:ts`, `relay`, and `next typegen`.

**`.github/instructions/*.instructions.md`** splits the monolithic root file into seven path-scoped files applied automatically by `applyTo` globs (backend, frontend, e2e, graphql, migrations, testing, ci). The root file drops to about 80 lines of genuinely shared context. This also feeds Copilot code review, so review comments become area-aware.

**`AGENTS.md`** is the portable cross-tool entrypoint (Codex, Cursor, Zed, Jules, etc.). It started as a fairly thin file, on the assumption that only non-Copilot tools would read it while Copilot itself only read `.github/copilot-instructions.md`. That assumption turned out to be wrong: Copilot's own code review and cloud coding agent now read `AGENTS.md` directly too, on top of `copilot-instructions.md`. So the two files had drifted into restating the same stack table, setup/dev/validation commands and critical rules — genuine same-tool duplication, not just cross-tool overlap. Resolved by making `AGENTS.md` the canonical source for that shared content and trimming `copilot-instructions.md` to defer to it, keeping only what's actually Copilot-specific (the path-scoped instructions table, the `hub-review` directive, data-flow/pitfalls notes). The one exception is Commit/PR conventions, which stays in `copilot-instructions.md` because that block is synced from an org-wide source (also mirrored in `CONTRIBUTING.md`) — `AGENTS.md` points to it there instead of risking a copy that falls out of sync.

**Custom agents reworked to stop restating rules.** `.github/agents/backend-code-writer.agent.md`, `frontend-code-writer.agent.md`, `reviewer.agent.md` and `commit-splitter.agent.md` used to duplicate stack/convention rules that also lived in the instructions files and in `.github/skills/*/SKILL.md` (in `commit-splitter`'s case, the full Conventional Commits format already canonical in `copilot-instructions.md`). They now explicitly defer: "follow skill X / instructions file Y for the rules, the text below only adds what's specific to this agent's posture." A rule that lives in one place can't silently drift into two disagreeing copies.

**React-query mandated over Relay for all new frontend work.** The frontend is mid-migration off Relay. `frontend.instructions.md`, `frontend-code-writer.agent.md`, `testing.instructions.md`, and the stack tables in `copilot-instructions.md`/`AGENTS.md` now consistently say react-query is mandatory for new data-fetching and Relay is only kept for existing pages, being phased out — including migrating a touched Relay component's data-fetching to react-query as part of an unrelated change, unless that's clearly out of scope.

**A `hub-review` skill (`.github/skills/hub-review/`) to keep the instruction surface honest going forward.** Rather than a one-off script, it is a small orchestrator (`SKILL.md`) that dispatches to four independent "lens" files under `references/`, each checking one specific kind of drift with its own evidence bar:

- **Stale reference** — a path, `yarn` command, `applyTo` glob, or hardcoded version that no longer resolves against the real repo.
- **Contradiction** — two authoritative files (instructions, agents, skills, `AGENTS.md`, `copilot-instructions.md`) giving conflicting guidance for the same situation.
- **Code-usage mismatch** — a documented convention the code no longer follows, or vice versa.
- **Duplication** — guidance restated across files instead of one linking to the other.

Findings that are objectively wrong (a broken path, a literal contradiction) get fixed directly. Findings that require a judgment call get escalated instead of guessed: a PR review comment when reviewing a diff, a direct question to the user in an interactive session, or a GitHub issue when unattended. `copilot-instructions.md` now points at it directly:

```
## Review
1. You must use the skill `hub-review`.
2. Follow the workflow.
```

**`.github/prompts/` was tried and then removed.** I initially added four task-scoped recipe files there (scaffold a backend module, a frontend page, a migration, write tests) and, separately, a mechanical drift-checking script wired to `yarn check:ai-instructions`. Both were dropped: the script was superseded by the `hub-review` skill described above, and the prompts directory added a fourth pillar with no reviewer of its own, so it was cut to keep the surface to three: instructions (path-scoped, automatic), skills (canonical rules and workflows), and agents (personas that defer to both).

## Real drift found and fixed along the way

Running the new setup steps and later the `hub-review` lenses against this repo (not guessing) turned up genuine, previously-unnoticed problems:

| Documented | Actual |
| --- | --- |
| Version `1.5.3` | `1.4.8` (now references `package.json`) |
| Node `24.13.1` | `.nvmrc` says `24.19.0` |
| Yarn `4.12.0` | `packageManager` says `4.18.0` |
| Next.js 15 | Next.js **16.3.1** |
| `.rules` at root | gitignored local editor file, not tracked |
| `yarn generate:enum` | script does not exist |
| `next lint` | `eslint .` |
| `.lintstagedrc.js` | `.lintstagedrc.cjs` |
| "lint-staged in backend then frontend" | one root run that dispatches per workspace |
| `apps/frontend/middleware.ts` proxies GraphQL/auth/document routes | that file has never existed; the real file is `proxy.ts` (Next.js 16 renamed the convention file) |
| react-query "preferred" for new data fetching (3 places) | left over from an earlier pass that only updated `frontend.instructions.md`; now "mandatory" everywhere, consistently |

Two CI bugs, fixed directly:
- **`dockerbuild-ci.yml` pinned `corepack prepare yarn@4.13.0`** while `package.json` declares `yarn@4.18.0`. Removed the pin; verified a bare `corepack enable` resolves 4.18.0 from `packageManager`.
- **`next-env.d.ts` is generated and gitignored**, so on a fresh checkout `yarn check-ts` fails with about twenty phantom `Cannot find module '@public/*.svg'` errors. `next typegen` resolves it; now both a setup step and a documented pitfall.

## Further fixes and conventions added after live review

- **The "no `console.log`" rule read as an absolute with only a backend remedy (`logApp`)**, but the frontend
  intentionally keeps three `console.log` call sites for dev-only GraphQL operation tracing
  (`src/lib/graphql-client.ts`, `src/lib/server-graphql-fetch.ts`, `src/relay/environment/fetch-fn.ts`), each gated
  behind `isDevelopment()` with an explicit `// eslint-disable-next-line no-console`. The docs were wrong, not the
  code: added a "Logging" section to `frontend.instructions.md` documenting the exception, and reworded the rule in
  `AGENTS.md` and `coding-conventions/SKILL.md` so it no longer implies these should be "fixed."
- **Comments:** `coding-conventions/SKILL.md` now says to keep comments to one line on *why* (not *what*), and to
  update an existing comment in place instead of leaving it stale and adding a new one alongside it.
- **Keeping the test suite lean:** `testing-validation/SKILL.md` now asks agents to actively spot tests their change
  makes redundant/obsolete/near-duplicate and propose consolidating or removing them (e.g. folding into an existing
  `it.each` dataset), rather than leaving bloat or adding a duplicate — still gated behind the existing "ask before
  touching an existing test" rule, so this only adds the proactive-detection step, not permission to silently rewrite.
- **Mutation check for new tests:** after writing or changing a test, briefly break the logic it's supposed to
  protect (comment out the line, invert a condition, remove a guard) and confirm the test fails, then revert. A test
  that still passes with the behavior broken isn't asserting anything real.
- **Bruno API collection was undocumented and silently going stale.** The repo has a real Bruno collection at
  `bruno/` (one folder per backend module, `.bru` files typed `graphql` or `http`), with nothing telling an agent to
  keep it in sync. Added an "API collection (Bruno)" section to `backend.instructions.md`: add or update the
  matching `.bru` request whenever a GraphQL operation or REST endpoint is added, renamed, or reshaped, matching the
  existing request shapes and reusing `bruno/environments/local.bru` variables. `graphql.instructions.md` points at
  it instead of restating it.
- **`commit-splitter.agent.md` fully restated the Conventional Commits format/types/signing rule** that's already
  canonical in `copilot-instructions.md`. Trimmed to a one-line pointer plus only what's specific to that agent
  (message clarity without branch context, imperative mood, no bracket prefixes).
- **`AGENTS.md` vs `copilot-instructions.md` duplication**, described above under "Approach".

## How to test

From a clean checkout:

```bash
corepack enable && yarn --version          # expect 4.18.0, not 4.13.0
yarn install --immutable
yarn workspace @xtm-hub/backend  generate:ts
yarn workspace @xtm-hub/frontend relay
yarn workspace @xtm-hub/frontend next typegen
yarn workspace @xtm-hub/frontend check-ts   # PASS
yarn workspace @xtm-hub/backend  check-ts   # PASS
```

I ran this sequence end to end on a clean tree; all steps succeed and both type-checks exit 0. Dropping the `next typegen` step reproduces the ~20 SVG errors, which is what justifies its inclusion. Every `yarn workspace ...` command and every backticked repo path across the new/changed files was verified to resolve against the actual `package.json` scripts and the filesystem.

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

No automated tests were added: this changes documentation, CI configuration, and agent instructions only, with no application code. The `copilot-setup-steps.yml` workflow is self-validating, since it runs on any push that touches it.

# Additional information

**Please add the `skip-feature-env` label.** This PR touches no application code, so a preview environment has no value here. I left the box unchecked rather than applying the label myself.

The documented CI pipeline was also out of date. It is now 8-way sharded with `detect-changes` and `run-lint` gates, and `ci.instructions.md` reflects the real job graph.

To stop this recurring, the instructions state a rule: do not hardcode versions in documentation, reference `.nvmrc`, `packageManager`, or the `catalog` block in `.yarnrc.yml` instead. The `hub-review` skill is the ongoing mechanism to catch the next round of drift instead of relying on a human noticing.

**Two things I deliberately left alone:**

- `test-feature-branch.yml` hardcodes Node `24.19.0`. Switching it to `node-version-file: .nvmrc` would break it, because that job has no `actions/checkout` step. Documented in `ci.instructions.md` instead.
- `.github/copilot-instructions.md` has a broken relative link to `LABELS.md`, but it sits inside the externally synced `<!-- filigran-conventions -->` block. Fixing it locally would be reverted on the next sync, so it likely needs fixing upstream.


.github/prompts/add-knex-migration.prompt.md
carinelebas, line 1: this may be duplicate with one of the skills ?

.github/instructions/migrations.instructions.md
carinelebas, line 1: this is again the same as one of the skills, and one of the prompts.. do we need these duplications?

jpkha: I remove the prompt to avoid duplications

.github/skills/coding-conventions/SKILL.md
```diff
@@ -25,6 +26,9 @@ stack-specific rules on top.
 - Consider performance implications (queries, loops, I/O, network calls,
   rerenders, overfetching, large lists) and avoid obvious bottlenecks.
 - Preserve existing naming conventions, folder structure, and import style.
-- Add concise comments only when the logic is non-obvious.
+- Add concise comments only when the logic is non-obvious, and keep them short — one line on *why*, not *what*.
+- When code you're touching already has a comment, update it in place instead of adding a new one alongside it.
+  A stale comment left next to a fresh one is worse than no comment; if the old one no longer applies, rewrite or
+  remove it rather than layering on more text.
```
Kroustille, line 32: 💯

Kroustille (approved):

.github/instructions/frontend.instructions.md
```diff
@@ -0,0 +1,146 @@
...
+soon as it enters the viewport (or on hover). Any `href` that resolves to a route with real side effects —
+for example paths under `/auth/*`, `/document/*`, `/user/picture`, or an App Router `route.ts` handler that
+authenticates, mutates data, or triggers a redirect with business logic (like
+`app/redirect/[identifier]/route.ts`) — must use `<Link href="..." prefetch={false}>`. Plain `<a href="...">`
+tags are unaffected and don't need this. When adding or reviewing any `<Link>`, check whether its target is
+a passive page/RSC route or a side-effecting endpoint before deciding on `prefetch`.
+
+## Internationalisation
+
+All user-facing strings go through `next-intl`. Add the key to `messages/en.json` and `messages/fr.json` — never
```
hervyt: **suggestion**: we can add the fact to always use `useTranslations() `> And then use the whole path `t('The.Beggining.Of.My.Path.Key') `; rather than `useTranslations('The.Beggining.Of.My.Path')` then `t('Key')`

hervyt (approved):

.github/instructions/frontend.instructions.md
carinelebas: we could also mention `messages/ja.json` (or all files in messages folder)

jpkha: Modification done
image

PR #3188, opened 2026-08-24
test(session): await on session cleanup to prevent flakiness (#3188)

Kroustille:
# Context
>

Some API tests are failing because we don't wait for session cleanup so it can be done after the test assertion.
This PR update the session cleanup function to return a promise and the tests await on that before asserting.

Nothing to test, only technical changes

## Related issues

No ticket.

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/backend/src/session-store-manager.test.ts
+++ b/apps/backend/src/session-store-manager.test.ts
@@ -73,27 +73,36 @@ describe('sessionStoreManager - Configuration-based Store Selection', () => {
       } as unknown as UserWithOrganizationsAndRole;
 
       // Should not throw any errors
-      expect(() => {
-        updateUserSession(testUser);
-      }).not.toThrow();
+      await expect(updateUserSession(testUser)).resolves.toBeUndefined();
     });
 
     describe('destroyUserSessions functionality', () => {
       it('should destroy all active sessions for the given user', async () => {
         const userId = 'destroy-target-' + uuidv4();
         const store = getSessionStoreInstance();
-        const allSpy = vi.spyOn(store, 'all').mockImplementation((callback) => {
-          callback(null, {
-            sessionA: { user: { id: userId } },
-            sessionB: { user: { id: 'another-user' } },
-            sessionC: { user: { id: userId } },
-          } as unknown as Record<string, SessionData>);
-        });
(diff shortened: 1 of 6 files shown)


apps/backend/src/utils/error/error.code.ts
```diff
@@ -97,7 +97,6 @@ export enum UnknownErrorCode {
   RemoveUserFromOrgaError = 'REMOVE_USER_FROM_ORGA_ERROR',
   RemoveUserFromPendingOrgaError = 'REMOVE_USER_FROM_PENDING_ORGA_ERROR',
   AcceptUserInPendingOrgaError = 'ACCEPT_USER_IN_PENDING_ORGA_ERROR',
-  EditUserSessionError = 'EDIT_USER_SESSION_ERROR',
```
jpkha, line 100: Why do you remove this error?  Should not be related to the issue?

Kroustille: I removed this throwing error because this way, the other sessions are cleaned up except the ones in error. A log is still written to know something bad happened.

What do you think ? 🙂

apps/backend/src/session-store-manager.test.ts
```diff
@@ -369,7 +402,7 @@ describe('sessionStoreManager - Configuration-based Store Selection', () => {
         last_name: 'Test',
         organizations: [],
         role: { id: 'role-id', name: 'USER' },
-      } as UserWithOrganizationsAndRole;
+      } as unknown as UserWithOrganizationsAndRole;
```
jpkha, line 405: What does property we are missing in this one to be force to cast as unknown ?

Kroustille: It's missing `roles_portal`, `selected_org_capabilities`, `salt`, `password`, `selected_organization_id`, `picture`, `disabled`, `last_login` which is quite a lot and that's why it was easier to force cast it to the right type.

I can have a base user with this properties populated with dummy values if it's better 😉

jpkha (approved):

PR #3195, opened 2026-08-25
feat(trialbundle): manage users page (#3123)

carinelebas:
# Context
>

Adds a "manage bundle trial users" page allowing organization admins to manage per-platform roles for users across the platforms bundled in an XTM platform trial
image
image

Notes: 
- there are some differences with the mock up linked to our filigran-ui components, like the 'selection' header appearing above the table or the or the table headers in uppercase.
- only try to add real users in auth0, otherwise an error will pop (since we cannot sync with auth0)

- Add users popup: add one or more users with roles across the bundle's actual platforms
image

- Remove users: single-user delete and bulk delete (multi-select) from the table; verify Auth0 group membership is actually revoked.
image

- Edit roles: bulk "group action" edit popup (multi-user, multi-platform) vs. the inline per-cell role select (single user/platform) — confirm the inline edit doesn't change roles on other platforms for that user or other users
image

- Bundle filtering: open manage-users for bundles with different platform combinations (e.g., missing OpenAEV) and confirm only relevant platforms/roles are shown everywhere (table, add form, edit form, role descriptions).
- Permissions: attempt to access the manage-users page as a non-admin/non-bypass user — should be blocked.
- XtmOne Auth0/group fix: activate a new XtmOne trial deployment and confirm its service group (Admin/User) is created; cancel an XtmOne deployment and confirm no Auth0 audience-deletion call/error occurs; confirm OpenAEV still doesn't get an Auth0 audience created, and OpenCTI still does.
- check light mode

## Related issues

- Related to #3123

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information
- (commits 1–6): incremental UI/data build-up — page skeleton → list users → add users → delete (single, then bulk) → tooltips
- Commit 7 is a pure refactor extracting shared form scaffolding with no behavior change, done specifically to avoid duplicating role-selection UI/logic when adding the edit flow in commit 8
- Commit 9 adds adds the inline single-cell role editor
- Commit 10 restricts the UI to only the platforms actually present in a given bundle
- Commit 11 locks the page down with proper permission checks.
- Commit 12 makes XtmOne's Auth0/service-group initialization for XtmOne trials.
- Commit 13 display/css updates
- commit 14: simply move files to the right folder


apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -22,24 +35,395 @@ import { ServiceGroupHelper } from './service-group.helper';
...
 export type UpdateGroupsPayload = { id: ServiceGroupId; userIds: UserId[] }[];
 
+const toServiceGroupResponse = (
+  serviceGroup: ServiceGroupModel
+): ServiceGroupResponse => ({
+  ...serviceGroup,
+  name: serviceGroup.name as ServiceGroupName,
+});
+
+const assertOrganizationAccess = async (
```
Kroustille: There is a lot of unexported functions in this app. Please move them into helpers, security helpers or domain objects.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -1076,6 +1076,10 @@ const sendDeploymentRequestCancelledNotifications = async (
 const deleteDeploymentRequestAudience = async (
```
Kroustille: You will probably have to update this function because I introduced `shouldDeleteDeploymentRequestAudience` in PR https://github.com/XTM-Hub/xtm-hub/pull/3164

apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -22,24 +35,395 @@ import { ServiceGroupHelper } from './service-group.helper';
...
+          },
+        })
+      )
+    );
+  } catch (error) {
+    logApp.error('Unable to send free_trial_user_added mail', { error });
+  }
+};
+
 export const ServiceGroupApp = {
```
Kroustille, line 23: In `addUsersToBundleGroups`, `removeUsersFromBundleGroups` and `updateBundleUserGroups`, there's no check on input user ids that they belong to the admin's organization.

Any user holding `ADMINISTRATE_ORGANIZATION` or `MANAGE_PLATFORM_REGISTRATION` in any organization can grant any Hub user Admin access on their OpenCTI/OpenAEV/XTMOne trial instance, and trigger free_trial_user_added mail to that user's address disclosing platform_url

apps/backend/src/modules/deployment/group/service-group.resolver.ts
```diff
@@ -19,6 +19,22 @@ const resolvers: Resolvers = {
...
+    },
+    bundleProducts: async (_, { serviceInstanceId }) => {
+      try {
+        return await ServiceGroupApp.loadBundleProducts(serviceInstanceId);
+      } catch (error) {
+        throw mapToGraphQLError(error);
+      }
+    },
   },
   Mutation: {
```
Kroustille, line 39: **not mandatory**: We may want to disable new mutations when `XTM_PLATFORM_TRIAL` feature flag is disabled.

apps/backend/src/utils/error/error.code.ts
```diff
@@ -129,6 +130,7 @@ export enum NotFoundErrorCode {
   OrganizationNotFound = 'ORGANIZATION_NOT_FOUND',
   UseCaseNotFound = 'USE_CASE_NOT_FOUND',
   SolutionCategoryNotFound = 'SOLUTION_CATEGORY_NOT_FOUND',
+  ServiceGroupNotFound = 'SERVICE_GROUP_NOT_FOUND',
```
Kroustille, line 145: This error codes does not have translations

apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -22,24 +35,395 @@ import { ServiceGroupHelper } from './service-group.helper';
...
+  if (!platformId || !platformIdentifier || newlyAddedUsers.length === 0) {
+    return;
+  }
+
+  try {
+    const platformConfiguration =
+      await PlatformConfigurationDomain.loadConfigurationByPlatform(platformId);
+    if (
+      !platformConfiguration ||
+      deploymentType !== DeploymentRequestDeploymentType.Trial ||
```
Kroustille: `deploymentType` can be checked before loading `platformConfiguration` to prevent querying DB when `deploymentType` is not trial

apps/backend/src/modules/deployment/group/service-group.resolver.ts
carinelebas: Currently, we don't check the feature flags on backend side. If we do this, we should also check that we cannot create bundle, cancel bundle etc to be coherent. This is not a logic we currently do in our devs 🤔

apps/backend/src/modules/deployment/deployment.app.ts
carinelebas: good point, I'll have a conflict at merge, and will manage it at that time.

apps/backend/src/modules/deployment/group/service-group.resolver.ts
Kroustille: We sometimes did it, sometimes not haha 😅 Currently if a bundle is created, does it have impact on platform and will it trigger instance creation ? 🤔

apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -27,19 +25,250 @@ export const ServiceGroupApp = {
...
+          newlyAddedUsers: users,
+          adminEmail: user.email,
+        });
+      })
+    );
+
+    return ServiceGroupApp.loadBundleUserServiceGroups(serviceInstanceId);
+  },
+
+  removeUsersFromBundleGroups: async (
```
Kroustille, line 174: I think we're missing the `ServiceGroupSecurityHelper.assertUsersBelongToOrganization` assertion here 🤔

Kroustille (approved):

apps/backend/src/modules/deployment/group/service-group.app.ts
carinelebas: I don't think it's needed to remove a user no? actually, if he doesn't belong to the orga, all the better if we remove it, no?

apps/backend/src/modules/deployment/group/service-group.resolver.ts
carinelebas: yes it will. It's currently only protected by the frontend. But I wonder if we want to restrict it on backend side, it could be useful if we need to make a deployment test without activating the FF...

Kroustille: Yeah you're right, we will need it for e2e testing at the end. LGTM then !

PR #3202, opened 2026-08-26
feat(trials): admin Dashboard & reordering  (#3104)

Julienger0:
# Context
>

Adds a new **Manage Trials** admin page that groups trials by bundle, so a bundle is managed as one
thing instead of one product at a time.
- One row per bundle, across four tabs (Waiting, Running, Expired, Cancelled), plus a Quotas tab.
- A **Products** column showing each product of the bundle as a coloured badge: yellow while it is
  starting, green when it is active, grey when it is over, with the exact status in the tooltip.
- Cleaner columns: readable names, `Jul 26, 2026` dates, connected status instead of registration
  status, and no more Status column.
- Actions apply to the whole bundle: cancelling cancels every product, and moving a bundle in the
  queue moves it as a block instead of shuffling products individually.
- Admins can change the bundle quota of a region. Raising it lets the next bundles in the queue
  start, products included
- Moving a bundle only reorders the other bundles of its region, and the priority shown is its rank
  in that region rather than a number shared by every region.
- The OpenCTI and OpenAEV dashboards now show the same table, so they gain the readable dates and
  the translated regions the bundle table already had, each keeping its own columns and sorting.
- Those dashboards no longer list the products that belong to a bundle, only the trials that were
  really requested on their own.

A few parts resemble the existing trial dashboards, especially the quota edit form. We chose not to
merge them into shared components: the two pages work differently underneath, and making one piece of
code serve both would add a lot of complexity for code that is meant to be deleted.

go to :
https://dev-pr-3202.hub.staging.filigran.io/app/admin/manage-trials

You can see the cancelled,expired, running , waiting and quotas tab
in waiting you can reorder per region
## Related issues

- Related to #3104

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/deployment/quota/deployment.quota.app.ts
```diff
@@ -166,6 +166,32 @@ export const DeploymentQuotaApp = {
...
+    newCapacity: number;
+    onRequestMoved: (request: DeploymentRequestModel) => Promise<void>;
+  }): Promise<void> => {
+    const key = bundleQuotaKey(region);
+
+    await DeploymentQuotaDomain.withLockedQuotaTransaction([key], async () => {
+      const { newAvailability } =
+        await DeploymentQuotaDomain.updateQuotaCapacity({ key, newCapacity });
+
+      for (let i = 0; i < newAvailability; i++) {
```
carinelebas: what if availability is negative? don't we do the same as for product quotas?

apps/frontend/graphql/deployment/deployment.query.graphql
carinelebas, line 1: we want to have only one file to manage queries on a domain (so it's easier to share fragments), not one per query. so this file should be named deployment.query.graphql

apps/frontend/messages/en.json
```diff
@@ -473,6 +473,45 @@
     "WelcomeTitle": "Welcome to XTM Hub",
     "WelcomeSubtitle": "Extend and scale your XTM Platform"
   },
+  "ManageTrials": {
```
carinelebas: is it not the same as what we have for the current dashboards? I don't think we should duplicate

apps/frontend/src/components/trials/manage/ManageTrialsList.tsx
carinelebas, line 1: why not reuse the already existing trial list tab? (or maybe update it if needed?)
It looks like you are duplicating components that are already existing.

apps/frontend/src/components/trials/manage/ManageTrialsTab.tsx
carinelebas, line 1: It looks like this file is also a more or less a duplication of what we already have

apps/frontend/graphql/deployment/deployment.mutation.graphql
carinelebas, line 1: same reason as for query, this file should be named deployment.mutation.graphql

apps/frontend/graphql/deployment/manage-trials-quotas.query.graphql
carinelebas, line 1: it should be in the query file

carinelebas (commented):
nice commit split! but why completely duplicate what already exists, especially the keys?

apps/frontend/src/components/trials/tab/quotas/TrialsTabQuotasPlatform.tsx
```diff
@@ -1,28 +1,21 @@
...
 import { ColumnDef } from '@tanstack/react-table';
 import { useTranslations } from 'next-intl';
 import { useMemo, useState } from 'react';
-import {
-  readInlineData,
-  useLazyLoadQuery,
-  useRefetchableFragment,
-} from 'react-relay';
+
+type QuotaRow = TrialsQuotaFragment & { id: string };
```
carinelebas: why not add the id in the fragment? if you need an id, you can as well use the real quota id?

apps/frontend/src/components/trials/tab/trials-tab.utils.ts
```diff
@@ -1,3 +1,72 @@
...
+  [DeploymentRequestHubStatus.Active]: SUCCESS_CLASS_NAME,
+  [DeploymentRequestHubStatus.Pending]: RUNNING_CLASS_NAME,
+  [DeploymentRequestHubStatus.Provisioning]: RUNNING_CLASS_NAME,
+  [DeploymentRequestHubStatus.Queued]: RUNNING_CLASS_NAME,
+  [DeploymentRequestHubStatus.Cancelled]: INACTIVE_CLASS_NAME,
+  [DeploymentRequestHubStatus.Expired]: INACTIVE_CLASS_NAME,
+  [DeploymentRequestHubStatus.Failed]: INACTIVE_CLASS_NAME,
+};
+
+export const resolveProductStatusClassName = (
```
carinelebas: question: is it really useful to have this configuration in this utils file? it's used only in one place and it's only configuration ?

apps/frontend/src/components/trials/tab/TrialsProductValues.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+import testRender from '@/utils/test/test-render';
+import {
+  DeploymentRequestHubStatus,
+  PlatformIdentifier,
+  TrialsProductFragment,
+} from '@graphql/generated';
+import { screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+
+const OPENCTI_PLATFORM_ID = 'opencti-tenant';
```
carinelebas: [nit] it can be confusing here to use 'tenant'. We have a tenant notion, and it's different from the plaform_id

apps/frontend/src/components/trials/tab/TrialsTab.tsx
```diff
@@ -32,495 +36,493 @@ import {
...
+  enableSorting,
+  cell: ({ row }: TrialsCellProps) => (
+    <span className="truncate">
+      {formatDate(row.original[id], 'DATE_MEDIUM') ?? '-'}
+    </span>
+  ),
+});
+
+const productColumns = (scope: TrialsScope, t: Translate): TrialsColumn[] => {
+  if (scope.kind === 'bundle') {
```
carinelebas: only the cell changes between bundle and product. cannot this code be factorized?

apps/frontend/src/components/trials/tab/TrialsTab.tsx
```diff
@@ -32,495 +36,493 @@ import {
...
+      {isReorderable && (
+        <>
+          <TooltipProvider>
+            <Tooltip>
+              <TooltipTrigger asChild>
+                <Button
+                  variant="tertiary"
+                  size="icon"
+                  className="border m-1"
+                  aria-label={t('TrialsDashboard.Actions.MoveToTop')}
```
carinelebas, line 251: we should use the same namespace for the keys in thie component: Either 'ManageTrials', or 'TrialsDashboard', but having both will be hard to maintain.

apps/frontend/src/components/trials/tab/TrialsProductValues.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface TrialsProductValuesProps {
+  products: readonly TrialsProductFragment[];
+  valueOf: (product: TrialsProductFragment) => string | null | undefined;
+}
+
+export const TrialsProductValues = ({
+  products,
+  valueOf,
+}: TrialsProductValuesProps) => {
+  const valuedProducts = sortProducts(products).filter(
```
carinelebas: maybe we should sort the products input, rather than sort them in each components? I think it would be easier to manage and ensure the order will be right even if we add a new component.

apps/frontend/src/components/trials/tab/TrialsTab.tsx
```diff
@@ -529,150 +531,105 @@ const TrialsTab = ({ type, platformIdentifier }: TrialsTabProps) => {
...
+    [data]
+  );
 
   const onSortingChange = (updater: unknown) => {
     handleSortingChange<DeploymentRequestOrdering>({
       updater,
       orderBy,
-      orderMode,
-      setOrderMode,
+      orderMode: orderMode === OrderingMode.Asc ? 'asc' : 'desc',
```
carinelebas: The sorting change function is used in other function. Is this change needed because we migrated out of relay? If so, it will be required in different other places in the future, it would be better to see if we can easily update handleSortingChange ? otherwise maybe in a follow up issue?

apps/frontend/src/components/trials/tab/quotas/TrialsTabQuotasPlatform.tsx
carinelebas, line 1: maybe it could be the opportunity to add tests for the quota components?

apps/frontend/src/components/trials/tab/TrialsTab.tsx
```diff
@@ -32,495 +37,530 @@ import {
...
+            isBundle
+              ? 'ManageTrials.Toast.BundleReordered'
+              : 'TrialsDashboard.Toast.TrialsReordered'
+          ),
+        });
+      },
+      onError,
+    });
+
+  const isBundleProduct = request.parent_id !== null;
```
carinelebas: if we fetch only 'standalone' or 'bundle' trials, we should not need to hide reorder or cancel button depending on if this is a bundle.

apps/frontend/src/components/trials/tab/trials-tab.utils.ts
Julienger0: You're right,Moved into TrialsProducts

apps/frontend/src/components/trials/tab/TrialsTab.tsx
Julienger0: done

apps/frontend/src/components/trials/tab/TrialsTab.tsx
Julienger0: Everything is under TrialsDashboard now

apps/frontend/src/components/trials/tab/TrialsTab.tsx
Julienger0: Yes, it comes from the react-query migration ,  I updated handleSortingChange

apps/frontend/src/components/trials/trials.const.ts
```diff
@@ -53,6 +53,8 @@ export const TRIALS_PRODUCT_ORDER: PlatformIdentifier[] = [
   PlatformIdentifier.Xtmone,
 ];
 
+export const STANDALONE_PARENT_ID_FILTER_VALUE = 'null';
```
carinelebas: question: do we need to have this value in the const file? it's used only trials-tab.utils.ts, I find it more complex to understand when we have the value in a const in a different file. Wdyt ?

carinelebas (approved):

Issue #3193, opened 2026-08-25
[perf] Improve loadServiceInstanceAndGrantAccess query


hervyt:
Since the query is getting too long on the production environment, consider rework it: 

<img width="719" height="49" alt="Image" src="https://github.com/user-attachments/assets/20924c7b-89bc-4311-905a-a3d75404ddc2" />

1. remove all the granting access on service access (the thing about subscription/user_services etc.) 
2. check in the frontend the way to get user_service_capabilities (for UPLOAD/DELETE/BYPASS capa) 
3. In the frontend check we still have access to all needed information about serviceInstance > Check we pass in the dataloader this time 😛 
4. In the DB, clean all subscriptions except Filigran on prod env => It will clean the user_services useless 
5. In the service admin page, check we can still grant subscriptio/user_service and capas to anyone

PR #3208, opened 2026-08-26
fix: improve loadServiceInstanceAndGrantAccess query (#3193)

hervyt:
# Context
> Optim the serviceInstanceById Query

## How to test
- You should see no regression

## Related issues

- Related to #3193 

# Checklist

## Quality

## Testing

### What tests have been made

- [X] Unit tests
- [X] Integration tests
- [ ] E2E tests
- [X] Local manual tests


apps/backend/src/migrations/20260826122300_remove_non_filigran_subscriptions.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  await knex('Subscription')
+    .whereNot('organization_id', 'ba091095-418f-4b4f-b150-6c9295e232c4')
+    .delete();
+}
```
jpkha: Are we sure about this ? Only Filigran needs subscription ? What about the trials ?

hervyt: OUhhhhhh 
Did not think about subscription's trials 😱

hervyt: Fixed

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
```diff
@@ -0,0 +1,47 @@
...
+  userServiceCapabilities: UserServiceCapability[];
+  subscription_id: SubscriptionId | null;
+};
+
+export const ServiceCapabilityApp = {
+  loadServiceCapabilitiesByServiceId: async (
+    serviceInstanceId: ServiceInstanceId
+  ): Promise<UserServiceCapabilitiesResponse> => {
+    const user = requestContext.requireUser();
+    const subscriptionPromise = user.selected_organization_id
```
jpkha: is it possible to have an user with a falsy selected_organization_id ?

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
```diff
@@ -0,0 +1,47 @@
...
+      ? await SubscriptionDomain.loadSubscriptionBy({
+          service_instance_id: serviceInstanceId,
+          organization_id: user.selected_organization_id,
+        })
+      : Promise.resolve(undefined);
```
jpkha: it seems weird to me to return a Promise and the other side a resolve of the Promise.
We should return a Promise or else the result of the promise, I think

apps/frontend/src/components/service/components/header/ServiceListHeaderButtons.tsx
```diff
@@ -41,11 +41,11 @@ const ServiceListHeaderButtons = ({}) => {
 
   return (
     <div className="flex gap-s">
-      {(canManageService || isBypass) && currentUserSubscriptionId && (
+      {(canManageService || isAdminOrga || isBypass) && subscriptionId && (
```
jpkha: Why do you add `isAdminOrga`, does `canManageService` already manage it ?

apps/frontend/app/(application)/app/(user)/service/opencti_custom_dashboards/[serviceInstanceId]/page.tsx
```diff
@@ -42,7 +42,7 @@ const Page = async ({ params }: ServiceCustomDashboardsPageProps) => {
           <PageLoader
             serviceInstance={
               response.data
-                .serviceInstanceByIdAndGrantAccess as unknown as serviceInstance_fragment$data
+                .serviceInstanceById as unknown as serviceInstance_fragment$data
```
Comment (jpkha) [inline apps/frontend/app/(application)/app/(user)/service/opencti_custom_dashboards/[serviceInstanceId]/page.tsx:45]: do we still need as unkown ?

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
```diff
@@ -0,0 +1,47 @@
...
+      ),
+      subscriptionPromise,
+    ]);
+
+    const userServiceCapabilities = serviceCapabilities.filter(
+      (
+        serviceCapability
+      ): serviceCapability is UserServiceCapabilityModel & {
+        user_service_id: string;
+      } => serviceCapability.user_service_id !== null
```
Kroustille: There's no need to apply this filter because `ServiceCapabilityDomain.loadServiceCapabilitiesByServiceId` returns a non nullable `user_service_id`

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
```diff
@@ -0,0 +1,47 @@
...
+  subscription_id: SubscriptionId | null;
+};
+
+export const ServiceCapabilityApp = {
+  loadServiceCapabilitiesByServiceId: async (
+    serviceInstanceId: ServiceInstanceId
+  ): Promise<UserServiceCapabilitiesResponse> => {
+    const user = requestContext.requireUser();
+    const subscriptionPromise = user.selected_organization_id
+      ? await SubscriptionDomain.loadSubscriptionBy({
```
Kroustille: The subscription is already joined and returned in with `ServiceCapabilityDomain.loadServiceCapabilitiesByServiceId` query. I'm not sure it's required to prefetch this subscription 🤔

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
hervyt: Nope, can't be null | undefined

apps/frontend/src/components/service/components/header/ServiceListHeaderButtons.tsx
hervyt: I rename the var, the canManageService is only the service capa

apps/frontend/src/hooks/use-service-capability.tsx
```diff
@@ -1,21 +1,78 @@
...
   capability: ServiceRestriction,
   serviceInstance?: serviceInstance_fragment$data
-) => {
+): boolean;
+function useServiceCapability(
+  capability: ServiceRestriction,
+  serviceInstance: serviceInstance_fragment$data | undefined,
+  options: { withSubscriptionId: true }
+): UseServiceCapabilityWithSubscriptionId;
+function useServiceCapability(
```
Kroustille: I find it confusing to have the same hook returning either a `boolean` or `UseServiceCapabilityWithSubscriptionId` depending on the given options 🤔

apps/backend/src/modules/security-management/service-capability/service-capability.resolver.ts
```diff
@@ -1,9 +1,23 @@
...
   ServiceCapabilityId:
     createRelayIdScalar<ServiceCapabilityId>('Service_Capability'),
+  Query: {
+    userServiceCapabilities: async (_, { service_instance_id }) => {
+      try {
+        return await ServiceCapabilityApp.loadServiceCapabilitiesByServiceId(
+          service_instance_id
+        );
+      } catch (error) {
+        throw mapToGraphQLError(error, UnknownErrorCode.AddCapabilitiesError);
```
Kroustille, line 18: The error does not seem to fit with the query

apps/backend/src/modules/security-management/service-capability/service-capability.domain.ts
```diff
@@ -20,6 +26,91 @@ export const ServiceCapabilityDomain = {
...
+        '=',
+        'Subscription_Capability.id'
+      )
+      .leftJoin(
+        'Service_Capability',
+        'Subscription_Capability.service_capability_id',
+        '=',
+        'Service_Capability.id'
+      )
+      .tap(restrictSubscriptionToUserOrganization)
```
Kroustille: This is a minor performance issue: using tap here adds a new join to subscription but we already have the subscription table joined in this query.
This tap can be replaced by `.andWhere('Subscription.organization_id', user.selected_organization_id)`

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
```diff
@@ -0,0 +1,47 @@
...
+          service_instance_id: serviceInstanceId,
+          organization_id: user.selected_organization_id,
+        })
+      : Promise.resolve(undefined);
+
+    const [serviceCapabilities, subscription] = await Promise.all([
```
Kroustille, line 23: This `Promise.all` is not useful because `subscriptionPromise` is already an awaited result. Requests still runs sequentially.

apps/frontend/app/(application)/app/(user)/service/opencti_custom_dashboards/[serviceInstanceId]/page.tsx
Comment (hervyt) [reply apps/frontend/app/(application)/app/(user)/service/opencti_custom_dashboards/[serviceInstanceId]/page.tsx:45]: Yes, otherwise we have a type error; maybe we can handle all these as unknown in a dedicated PR everywhere ?

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
hervyt: It is in the case the user is only AdminOrga and does not have serviceCapabilities + loadServiceCapabilitiesByServiceId should not return this subscription_id anymore 🧑‍🎤  I remove it 📦

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
hervyt: Now it's 2 promises :)

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
hervyt: Promise.all is still useful here: subscriptionPromise is not awaited before Promise.all, so both async calls start first and are then awaited together.
 They run concurrently, not sequentially

Kroustille: Haha I saw an `await` when I posted the comment 😛 Looks good now 👍

apps/backend/src/modules/security-management/service-capability/service-capability.app.ts
Kroustille: If it can't be `null | undefined`, it's not required to use the ternary to have `subscriptionPromise`

apps/backend/src/modules/security-management/service-capability/service-capability.app.test.ts
```diff
@@ -0,0 +1,37 @@
...
+describe('serviceCapability app', () => {
+  const serviceInstanceId = SERVICES.INSTANCES.EPIC.ID;
+  const userId = TEST_ORGANIZATIONS.SECOND_ORGANIZATION.USERS.SIMPLE.ID;
+  const selectedOrganizationId = TEST_ORGANIZATIONS.SECOND_ORGANIZATION.ID;
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('should propagate domain error when loading capabilities fails', async () => {
```
Kroustille, line 81: This tests should assert the different use cases in the app fucntion and by calling DB (integration test)

Kroustille (approved):

PR #3209, opened 2026-08-26
feat(trials): active details page (#3102)

theochhean:
# Context

This PR introduces the XTM Platform Trial bundle page, giving users a single place to view and access all products included in their active XTM Platform trial (OpenCTI, OpenAEV, and XTM One)
- New `activeXtmPlatformBundle`GraphQL query exposing the active bundle and its child products
- A dedicated bundle page at `/app/service/xtm-platform-trial/{serviceInstanceId}`, rendering a product card per product (name, connection status, last connection, roles, and an Access button)
- A left-menu CTA entry 
- `Manage users` now links to `/app/service/xtm-platform-trial/{serviceInstanceId}/manage-users `and `See more` (trial guide) links to `/app/service/xtm-platform-trial-guide` (both target pages are delivered in a separate chunk)
- Refactor from `formatDate` to `useDateFormatter`

## How to test

https://github.com/user-attachments/assets/9c83f4e6-7d0c-49af-94c3-2a2a74800b3a

- Add rights to your account for products trial (Settings > OpenCTI / OpenAEV Trial > Running)
- From the left menu, click the XTM Platform Trial CTA
- Verify the page renders one card per product if present in the bundle (OpenCTI, OpenAEV, XTM One) with the correct name, connection status, last connection date, and roles
- Verify the Acces btns are available for OpenCTI / Open AEV (if you added permissions)
- Verify you can update the product name
- Check that redirect links are right (trial guide and manage users links to non existing page for now but verify url)
- XTM One cannot be tested on feature env, will be tested on staging

Here's how it looks in local with mocked xtm one call (+ disconnected product card)
Capture d’écran 2026-08-28 à 14 38 14

- Light theme : waiting for design answer

## Related issues

- Related to #3102

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -556,6 +559,76 @@ export const DeploymentApp = {
...
+      return null;
+    }
+
+    const children = await DeploymentRequestDomain.loadDeploymentRequestsBy({
+      parent_id: bundle.id,
+    });
+
+    const productData = await Promise.all(
+      children
+        .filter(
```
carinelebas: is this filter needed? children always have a PlatformIdentifier. If it happens that there is not, we have a big issue, and it should be an error, not silently ignored.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -556,6 +559,76 @@ export const DeploymentApp = {
...
+            );
+          const childServiceInstance =
+            await ServiceInstanceDomain.loadServiceInstanceBy({
+              id: child.service_instance_id,
+            });
+          return { child, configuration, roles, childServiceInstance };
+        })
+    );
+
+    const license =
```
carinelebas: The license for the bundle will always be trial. I don't think there is any need to look on the products? 🤔 
If we were to pick one from the products, then we'll have to define a logic, here we pick a randome one?

apps/frontend/graphql/xtm-platform-bundle/active-xtm-platform-bundle.query.graphql
carinelebas, line 1: since we are fetching deployments, I think this file should be in deployment/deployment.query.graphql, and not in it's dedicated folder. This can allow to build and use grapqhl fragment more easily

apps/frontend/src/components/registration/PlatformIdentifierMapping.tsx
```diff
@@ -41,14 +41,13 @@ export const PlatformMetadataMapping: Record<
     docUrl: 'https://docs.openaev.io/latest/administration/hub/',
     Icon: ({ className }) => <OpenAevIconIcon className={className} />,
   },
-  // Placeholder: xtmone has no trial UI yet, replace when built.
```
carinelebas, line 47: 💯 I've seen the same change in other PR, you are the only one removing the comment !

apps/frontend/src/components/xtm-platform-trial/BundleInfoCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+  </div>
+);
+
+export const BundleInfoCard = ({
+  bundle,
+  canManage,
+  organizationId,
+}: BundleInfoCardProps) => {
+  const t = useTranslations();
+  const tBundle = useTranslations('XtmPlatformTrial');
```
carinelebas: we should not use scoped translations, but simply `useTranslations()`

apps/frontend/src/components/xtm-platform-trial/BundleProductCard.tsx
```diff
@@ -0,0 +1,185 @@
...
+  [PlatformIdentifier.Openaev]: 'text-filigran-brand-primary',
+  [PlatformIdentifier.Xtmone]: 'text-filigran-ia-main',
+};
+
+export const BundleProductCard = ({
+  product,
+  xtmoneStatus,
+  canManage,
+}: BundleProductCardProps) => {
+  const tProducts = useTranslations('XtmPlatformTrial.Products');
```
carinelebas: we should use directly `useTranslations()`

apps/frontend/src/components/xtm-platform-trial/BundleProductCard.tsx
```diff
@@ -0,0 +1,185 @@
...
+import Link from 'next/link';
+import { useState } from 'react';
+
+interface BundleProductCardProps {
+  product: XtmPlatformBundleProduct;
+  xtmoneStatus: XtmoneStatusState;
+  canManage: boolean;
+}
+
+const PRODUCT_LOGO_COLOR: Record<PlatformIdentifier, string> = {
```
carinelebas: maybe this should be in PlatformIdentifierMapping?

apps/frontend/src/components/xtm-platform-trial/xtm-platform-bundle.types.ts
```diff
@@ -0,0 +1,7 @@
+import { ActiveXtmPlatformBundleQuery } from '@graphql/generated';
+
+export type XtmPlatformBundleData = NonNullable<
+  ActiveXtmPlatformBundleQuery['activeXtmPlatformBundle']
+>;
+
+export type XtmPlatformBundleProduct = XtmPlatformBundleData['products'][number];
```
carinelebas: question: maybe we should use a graphql fragment instead of this type?

apps/frontend/src/components/xtm-platform-trial/XtmoneConnectionStatus.tsx
```diff
@@ -0,0 +1,97 @@
...
+const StatusIcon = ({
+  entry,
+  isLoading,
+  unavailable,
+}: {
+  entry?: XtmoneIntegrationStatusEntry;
+  isLoading: boolean;
+  unavailable: boolean;
+}) => {
+  const tProducts = useTranslations('XtmPlatformTrial.Products');
```
carinelebas: we should use directly `useTranslations()`

apps/frontend/src/components/xtm-platform-trial/XtmPlatformTrialPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+  PlatformIdentifier,
+  PortalCapability,
+  useActiveXtmPlatformBundleQuery,
+} from '@graphql/generated';
+import { xtmPlatformBundleKeys } from '@graphql/xtm-platform-bundle/xtm-platform-bundle.keys';
+import { useTranslations } from 'next-intl';
+import { useContext } from 'react';
+
+export const XtmPlatformTrialPage = () => {
+  const tBundle = useTranslations('XtmPlatformTrial');
```
carinelebas: we should use directly `useTranslations()`

apps/frontend/src/components/xtm-platform-trial/XtmPlatformTrialPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const productOrder: Record<string, number> = {
+    [PlatformIdentifier.Opencti]: 0,
+    [PlatformIdentifier.Openaev]: 1,
+    [PlatformIdentifier.Xtmone]: 2,
+  };
+  const orderedProducts = [...bundle.products].sort(
+    (a, b) =>
+      (productOrder[a.platform_identifier] ?? Number.MAX_SAFE_INTEGER) -
+      (productOrder[b.platform_identifier] ?? Number.MAX_SAFE_INTEGER)
+  );
```
carinelebas: can't we simply hardcode the sorted list? it would be easier?

apps/frontend/src/components/xtm-platform-trial/BundleProductCard.tsx
```diff
@@ -28,10 +32,10 @@ interface BundleProductCardProps {
   canManage: boolean;
 }
 
-const PRODUCT_LOGO_COLOR: Record<PlatformIdentifier, string> = {
-  [PlatformIdentifier.Opencti]: 'text-filigran-brand-primary',
-  [PlatformIdentifier.Openaev]: 'text-filigran-brand-primary',
-  [PlatformIdentifier.Xtmone]: 'text-filigran-ia-main',
+const PRODUCT_TEXT_LOGO: Record<PlatformIdentifier, StaticImageData> = {
```
carinelebas: This should be in `PlatformIdentifierMapping`

apps/frontend/src/components/xtm-platform-trial/TrialLimitationsCard.tsx
carinelebas, line 1: we should reuse [XtmPlatformTrialLimitations component ](https://github.com/XTM-Hub/xtm-hub/blob/2524e0d927206a62925e178437c492dd7e541e4f/apps/frontend/src/components/service/trial-instances/xtm-platform-trial/XtmPlatformTrialLimitations.tsx)

apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/page-loader.tsx
```diff
@@ -14,11 +14,11 @@ const breadcrumbValue = [
   },
 ];
 
-const PageLoader = () => {
+const PageLoader = ({ serviceInstanceId }: { serviceInstanceId: string }) => {
```
Comment (carinelebas) [inline apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/page-loader.tsx:17]: question: why do you need to change this path and add the serviceInstanceId in it?

apps/backend/src/modules/deployment/deployment.graphql
```diff
@@ -235,6 +235,27 @@ type DeploymentAvailability {
...
+  platform_identifier: PlatformIdentifier!
+  service_instance_id: ServiceInstanceId!
+  name: String
+  connectivity_status: PlatformConfigurationStatus
+  last_connectivity_check: Date
+  roles: [ServiceGroup!]!
+  url: String
+}
+
+type XtmPlatformBundle {
```
carinelebas: I think we should use DeploymentRequest type, which already extends node. We can talk about it.

apps/frontend/src/components/xtm-platform-trial/XtmPlatformTrialPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const canManage =
+    hasCapability?.(PortalCapability.Bypass) ||
+    hasOrganizationCapability?.(
+      OrganizationCapability.AdministrateOrganization
+    ) ||
+    hasOrganizationCapability?.(
+      OrganizationCapability.ManagePlatformRegistration
+    ) ||
+    false;
```
Kroustille: You can directly use the hooks `useAdminByPass` and `useGranted` here.

apps/frontend/messages/en.json
```diff
@@ -64,6 +64,49 @@
...
+      "Role": "Role",
+      "NoRole": "None.",
+      "NoRoleHint": "Please contact the trial requester.",
+      "StatusActive": "Connected",
+      "StatusNotConnected": "Not connected",
+      "AccessProduct": "Access {productName}",
+      "StatusUnavailable": "Status unavailable",
+      "ProductName": "Product name",
+      "ConnectionOpencti": "Connection OpenCTI",
+      "ConnectionOpenaev": "Connection OpenAEV",
```
Kroustille: This 2 translations can be one parametrized translation and `OpenCTI` or `OpenAEV` can come from `PlatformMetadataMapping`

apps/frontend/src/components/registration/PlatformIdentifierMapping.tsx
Kroustille: Yes haha I only changed the icon in my trial guide PR but I didn't see the comment 😅

apps/frontend/messages/en.json
```diff
@@ -64,6 +64,49 @@
...
+    "Title": "Welcome to your XTM Platform Trial",
+    "Subtitle": "Everything you need for your trial is right here: your trial details and time remaining, direct access to your products, and a direct line to the Filigran team whenever you need support or have questions.",
+    "Guide": {
+      "Title": "XTM Platform Trial Guide",
+      "Content": "Explore what we're building, what's coming next, and where the XTM Platform is headed. Full transparency. No surprises.",
+      "SeeMore": "See more"
+    },
+    "Limitations": {
+      "Title": "Trial limitations",
+      "Intro": "The XTM platform free trial comes includes all Enterprise Edition SaaS features, but with limited computing power and storage:",
```
Kroustille: This sentence is not gramatically correct (**comes includes**)

apps/frontend/src/components/xtm-platform-trial/useXtmoneIntegrationStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface XtmoneStatusState {
+  data?: XtmoneIntegrationStatus;
+  isLoading: boolean;
+  isError: boolean;
+  hasUrl: boolean;
+}
+
+const XTMONE_STATUS_TIMEOUT_MS = 8000;
+
+const fetchXtmoneIntegrationStatus = async (
```
Kroustille: I'm wondering if it's a good idea to fetch status from frontend, or use next backend or backend app as a proxy to fetch status 🤔 
What do you think @carinelebas ?

carinelebas: Since we are fetching data from another "backend" (xtmone), I don't think this should be done in the same query as the one displaying the main screen. In case it takes time to answer, or doesn't respond, we may delay the display of the main screen, which I don't think is something we want. And in this case, I'm not sure about the added value of using the backend as proxy ? Wdyt ?

Kroustille (approved):

apps/frontend/src/components/xtm-platform-trial/useXtmoneIntegrationStatus.ts
Kroustille: If at any moment the XTM One backend enable CORS headers without specifying the Hub URL as a potential caller, this fetch will cease to work.
In any case if the platform url is updated with a malicious URL, any web browser will fetch that page and it can lead to XSS attacks 😬 
I think in general it's safer to call another domain backend from our own backend instead of the frontend.

Kroustille (commented):
Approved too fast, agreed with Carine comments, needs to fix it before approval

apps/frontend/src/components/xtm-platform-trial/useXtmoneIntegrationStatus.ts
carinelebas: ok, maybe we can have then a small dedicated passthrough endpoint!

apps/backend/src/modules/deployment/deployment.graphql
```diff
@@ -142,6 +142,12 @@ type DeploymentRequest implements Node {
   platform_url: String
   parent_id: DeploymentRequestId
   url: String
+  name: String
+  connectivity_status: PlatformConfigurationStatus
+  last_connectivity_check: Date
+  roles: [ServiceGroup!]
+  license: PlatformContract
```
carinelebas: we should not add these fields to the deploymentRequest. These are already in the registeredPlatform object, we should reuse it. 
We should simply add registeredPlaform in the deployment request, and have a field resolver that will call `RegistrationDomain.loadRegisteredPlatform(service_instance_id)`

Note: when adding fields to a graphql type, we should be careful that these fields can be returned for all queries/mutation that use them. Using field resolvers allows to do this easily. These fields that you added were only available when using the new mutation. This makes the code harder to maintain if you don't know with which query graphql fields are provided or not.

apps/frontend/src/components/xtm-platform-trial/useXtmoneIntegrationStatus.ts
```diff
@@ -1,23 +1,16 @@
...
-  opencti: XtmoneIntegrationStatusEntry;
-  openaev: XtmoneIntegrationStatusEntry;
-  linked: boolean;
-  last_checked_at: string | null;
-}
-
-interface XtmonePlatformConfigResponse {
-  integration_status?: XtmoneIntegrationStatus;
-}
+export type XtmoneIntegrationStatusEntry = XtmoneIntegrationStatus['opencti'];
```
carinelebas: you can avoid defining this type (and XtmoneIntegrationStatus) by using graphql fragments. 
I think it would be better than defining  XtmoneIntegrationStatusEntry only based on 'opencti'

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -93,6 +94,58 @@ export const XTM_PLATFORM_BUNDLE_SERVICE_INSTANCE_NAME = 'XTM Platform Bundle';
...
+  try {
+    parsedUrl = new URL(baseUrl);
+  } catch {
+    return null;
+  }
+  if (parsedUrl.protocol !== 'https:' && parsedUrl.protocol !== 'http:') {
+    return null;
+  }
+
+  const url = `${baseUrl.replace(/\/$/, '')}/api/v1/platform/config`;
```
carinelebas: it's better to use `new URL('/api/v1/platform/config', parsedUrl)` , which also correctly handles a  baseUrl  that already includes a path/query and avoids the ad hoc trailing-slash regex

apps/backend/src/modules/deployment/deployment.graphql
```diff
@@ -142,6 +142,12 @@ type DeploymentRequest implements Node {
   platform_url: String
   parent_id: DeploymentRequestId
   url: String
+  name: String
```
carinelebas: for the name, we should add a field_resolver that will return the service_instance_id object, and have the front end only query the name.

apps/frontend/src/components/menu/navigation/private/use-private-navigation.ts
carinelebas, line 1: I find it strange that you need to do format commits? you don't have husky commit hook installed?

apps/backend/src/modules/deployment/deployment.resolver.ts
```diff
@@ -4,17 +4,39 @@ import {
...
+      ),
+    service_instance: async ({ service_instance_id }) => {
+      const serviceInstance = await ServiceInstanceDomain.loadServiceInstanceBy(
+        { id: service_instance_id as ServiceInstanceId }
+      );
+      return serviceInstance
+        ? (serviceInstance as unknown as ServiceInstance)
+        : null;
+    },
+    children: ({ id }) =>
```
carinelebas: you will have a conflict with this: https://github.com/XTM-Hub/xtm-hub/blob/61ddf67be13e477dd339efe361060bce669c5355/apps/backend/src/modules/deployment/deployment.resolver.ts#L19

apps/backend/src/modules/deployment/deployment.resolver.ts
```diff
@@ -76,6 +98,36 @@ const resolvers: Resolvers = {
         );
       }
     },
+    activeXtmPlatformBundle: async (_, { serviceInstanceId }, context) => {
+      try {
+        return await DeploymentApp.loadActiveXtmPlatformBundle(
+          context.user,
```
carinelebas: for consistency with the other functions, you should rather call `const user = requestContext.requireUser();` in the app, than pass the user as parameter. 
it's the same value. The main idea is that when we want to get the user making the query, we get it from the context, and don't pass it as parameter, unless it has a real meaning in the function behavior (update user for example).

apps/frontend/src/components/menu/navigation/private/use-private-navigation.ts
```diff
@@ -468,6 +486,19 @@ export const usePrivateNavigation = (): NavigationConfig => {
       label: tMenu('Slack'),
       external: true,
     },
+    ...(isXtmPlatformBundleEnabled && xtmPlatformBundleServiceInstanceId
```
carinelebas: addition of this button was initially planned in another chunk. It should be added whatever the status of the trial, so the addition should not be linked to the existence of an active trial. 
We don't need to make the query in this file, it could only be done when opening the page. This button should link to xtm-platform-trial page, which will display different contents depending on the existence/status of the trial.

apps/backend/src/modules/deployment/deployment.app.ts
```diff
@@ -91,6 +92,58 @@ export const XTM_PLATFORM_BUNDLE_SERVICE_INSTANCE_NAME = 'XTM Platform Bundle';
 
 const DEPLOYMENT_REQUEST_LOCK_NAMESPACE = 'deployment_request';
 
+const XTMONE_STATUS_TIMEOUT_MS = 8000;
+
+interface XtmonePlatformConfigResponse {
+  integration_status?: XtmoneIntegrationStatus;
+}
+
+const fetchXtmoneIntegrationStatus = async (
```
Kroustille: This could be moved in the thirdparty folder 🤔

Kroustille (approved):

apps/backend/src/thirdparty/xtmone/xtmone.ts
carinelebas, line 1: [nit] it would be nice to have some tests

carinelebas (approved):

PR #3222, opened 2026-08-27
feat(connectors): exposing registeredProductVersions GraphQL query (#3203)

jpkha:
# Context

This PR exposes the list of product versions that have been registered for a platform (OpenCTI, OpenAEV, ...), through two new read paths:

- **GraphQL query** `registeredProductVersions(product: PlatformIdentifier!): [RegisteredProductVersion!]!`, returning the registered `version` and `created_at` for the given product, ordered by version (descending).
- **REST endpoint** `GET /:product/versions`, returning the same data as JSON (`{ product, versions: [{ version, created_at }] }`) for consumers that cannot use GraphQL (e.g. connectors, external scripts). It is protected by a per-IP rate limiter (300 req/min) reusing the shared `ip-rate-limit.util.ts` helper extracted from the manifest endpoint.

Versions are now persisted in a new `ProductVersion` table (added via migration) whenever `newProductVersion` is called, instead of only triggering manifest generation. Both the query and the REST endpoint read from this table.

## How to test

- GraphQL: run the `registeredProductVersions(product: OPENCTI)` query (see `bruno/manage-product-version/RegisteredProductVersions.bru`).
- REST: call `GET /opencti/versions` (see `bruno/manage-product-version/GetRegisteredProductVersions.bru`) and check the JSON response.
- Register a new version through the `newProductVersion` mutation, then confirm it shows up in both the query and the REST endpoint.

## Related issues

- Related to #3203 

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/modules/manage-product-version/manage-product-version.resolver.ts
```diff
@@ -10,6 +17,13 @@ const resolvers: Resolvers = {
         type: ManifestType.Connector,
       });
 
+      await ManageProductVersionDomain.registerProductVersion({
```
Kroustille: I think it's the responsibility of `registerProductVersion` to validate and format version to a `padded_version` before inserting it.

Kroustille: Resolver should not call domain directly but call app layer

Kroustille (approved):

PR #3226, opened 2026-08-28
feat(connectors): display only latest connectors (#3204)

jpkha:
# Context
This PR adds a `DECOUPLING_CONNECTORS` feature flag that, when enabled, restricts the connector library listing to only show **connectors** tagged with both `latest` and `decoupling`. Connectors missing either tag (decoupled-only, or legacy/untagged connectors) are hidden from the listing. Other integration subtypes (feeds, dashboards, etc.) are not affected and keep the previous `excludeDecouplingTag` filtering behaviour regardless of the flag.

Key changes:
- Added `DECOUPLING_CONNECTORS` to the `FeatureFlag` enum (backend GraphQL schema + generated types, and the corresponding frontend generated types/mocks).
- Added `restrictConnectorsToLatestDecoupling` filtering logic in `document.domain.ts`, applied only to the integrations listing query (`OPENCTI_INTEGRATION_DOCUMENT_TYPE`) when the feature flag is enabled; falls back to the existing `excludeDecouplingTag` behaviour otherwise.
- Added unit tests covering the new behaviour: only connectors tagged with both `latest` and `decoupling` are returned when the flag is on, while other integration types remain untouched.
- Updated the `IngestManifestFragments.bru` sample payload (bumped connector version, added the required `solution_categories` field) and documented the new/optional manifest fragment fields.

- Create a new connector by calling IngestManifestFragments
- Check that you can see the new connector
- You should not see the old connector

## Related issues

- Related to #3204

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information

### DIFF
--- a/apps/backend/src/modules/document/domain/document.domain.ts
+++ b/apps/backend/src/modules/document/domain/document.domain.ts
@@ -3,6 +3,7 @@ import { db, dbRaw, paginate } from '../../../../knexfile';
 import {
   DocumentConnection,
   DocumentMetadataKeyCode,
+  FeatureFlag,
   IntegrationType,
   Organization,
   QueryDocumentsArgs,
@@ -17,11 +18,13 @@ import {
 import { ServiceInstanceId } from '../../../model/kanel/public/ServiceInstance';
 import User, { UserId } from '../../../model/kanel/public/User';
 import { UnknownErrorCode } from '../../../utils/error/error.code';
+import { isFeatureEnabled } from '../../../utils/feature-flag.util';
 import { omit } from '../../../utils/utils';
 import { isLtsVersion } from '../../../utils/versioning';
 import {
   ConnectorV2,
   INTEGRATION_CONNECTOR_V2_METADATA_KEYS,
+  OPENCTI_INTEGRATION_DOCUMENT_TYPE,
 } from '../../shareable-resource/opencti/integration/integration.model';
 import { Document, DOCUMENT_TYPE, WithDocumentId } from '../document.helper';
 
@@ -45,6 +48,7 @@ import { stripNulls } from '../../../utils/typescript';
 import {
   ManifestFragmentHelper,
   TAG_DECOUPLING,
+  TAG_LATEST,
 } from '../../shareable-resource/manifest-fragment/manifest-fragment.helper';
 import { isUserRestrictedToActiveDocument } from '../document.security';
 import {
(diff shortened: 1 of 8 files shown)


apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -45,6 +48,7 @@ import { stripNulls } from '../../../utils/typescript';
 import {
   ManifestFragmentHelper,
   TAG_DECOUPLING,
+  TAG_LATEST,
 } from '../../shareable-resource/manifest-fragment/manifest-fragment.helper';
```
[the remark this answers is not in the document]
jpkha: I only want to restrict to TAG_LATEST for the moment

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -62,6 +66,40 @@ const excludeDecouplingTag = (query: Knex.QueryBuilder) =>
...
+    .select(dbRaw('1'))
+    .from('Document_Metadata')
+    .whereRaw('"Document_Metadata"."document_id" = "Document"."id"')
+    .andWhere('Document_Metadata.key', DocumentMetadataKeyCode.IntegrationType)
+    .andWhere('Document_Metadata.value', IntegrationType.Connector);
+
+// For connector documents, only keep those tagged with both TAG_LATEST and TAG_DECOUPLING
+// (case-insensitive). Other integration subtypes keep the default excludeDecouplingTag behaviour.
+// Used behind the DECOUPLING_CONNECTORS feature flag.
+const restrictConnectorsToLatestDecoupling = (query: Knex.QueryBuilder) =>
```
Kroustille, line 71: Instead of calling buildIsConnectorSubquery twice (once negated, once not) inside an OR, compute the "is connector" fact once and branch on it:
- Add a single `LEFT JOIN` to Document_Metadata (aliased, e.g. `dm_integration_type`) filtered in the join's ON clause to `key = IntegrationType`, instead of a WHERE-based correlated subquery. This turns "is connector" into a nullable column (dm_integration_type.value) that Postgres computes once per row during the join, not twice via re-executed subqueries.
- Then express the OR purely in terms of that joined column:
  - `WHERE (dm_integration_type.value IS DISTINCT FROM 'connector' AND <excludeDecouplingTag>) OR (dm_integration_type.value = 'connector' AND <latest+decoupling tags>)`
- This mirrors the existing `loadDocumentsByMetadata` convention (`LEFT JOIN Document_Metadata + groupBy`) already used elsewhere in the same file, removing the stylistic inconsistency and the double evaluation in one move.
- Keep the existing `groupBy(['Document.id'])` (already present) to collapse the join's row multiplication back to one row per document: verify no duplicate rows are introduced for documents with multiple metadata rows sharing the same key (shouldn't happen given the (document_id, key) primary key, but worth a one-line comment noting why the join is safe).

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -62,6 +66,40 @@ const excludeDecouplingTag = (query: Knex.QueryBuilder) =>
...
+      .orWhere((connectorBuilder) => {
+        connectorBuilder
+          .whereExists(buildIsConnectorSubquery)
+          .whereRaw(
+            `? ILIKE ANY(COALESCE("Document"."tags", ARRAY[]::text[]))`,
+            [TAG_LATEST]
+          )
+          .whereRaw(
+            `? ILIKE ANY(COALESCE("Document"."tags", ARRAY[]::text[]))`,
+            [TAG_DECOUPLING]
```
Kroustille: This `ILIKE ANY` are non indexable. It could be replaced and combined in one check with `"Document"."tags" @> array['latest','decoupling']`

apps/backend/src/modules/document/domain/document.domain.ts
```diff
@@ -292,7 +330,18 @@ export const DocumentDomain = {
...
+          field['Document.type'] === OPENCTI_INTEGRATION_DOCUMENT_TYPE;
+        if (
+          isIntegrationsListing &&
+          isFeatureEnabled(FeatureFlag.DecouplingConnectors)
+        ) {
+          restrictConnectorsToLatestDecoupling(qb);
+        } else {
+          excludeDecouplingTag(qb);
+        }
+      });
```
Kroustille: This filter is only applied to `loadDocuments` query but not at other places where `excludeDecouplingTag` is currently applied

apps/backend/src/modules/document/domain/document.domain.ts
jpkha: interesting, did not know that

apps/backend/src/modules/document/domain/document.domain.ts
jpkha: Here the new query, WDYT
```
query
    // LEFT JOIN instead of two EXISTS subqueries (one negated, one not): computes
    // "is connector" once per row. Safe from row duplication since (document_id, key)
    // is Document_Metadata's primary key, so at most one row matches per document.
    .leftJoin({ dm_integration_type: 'Document_Metadata' }, function () {
      this.on('dm_integration_type.document_id', '=', 'Document.id').andOnVal(
        'dm_integration_type.key',
        DocumentMetadataKeyCode.IntegrationType
      );
    })
    .where((outer) => {
      outer
        .where((nonConnectorBuilder) => {
          nonConnectorBuilder
            .whereRaw('"dm_integration_type"."value" IS DISTINCT FROM ?', [
              IntegrationType.Connector,
            ])
            .modify(excludeDecouplingTag);
        })
        .orWhere((connectorBuilder) => {
          connectorBuilder
            .where('dm_integration_type.value', IntegrationType.Connector)
            .whereRaw(`"Document"."tags" @> ARRAY[?, ?]::text[]`, [
              TAG_LATEST,
              TAG_DECOUPLING,
            ]);
        });
    });
```

Kroustille: Looks good to me !

Kroustille (approved):

PR #3265, opened 2026-09-02
[Pause] feat(integrations): add OpenCTI version filters (#3205)

jpkha:
# Context

Connectors on OpenCTI integration pages (private "My services" and public "Cybersecurity solutions") now support filtering by OpenCTI platform version, so users can quickly find connectors compatible with the OpenCTI version they run.

- Adds a new `OpenctiVersionFilter` component (single-select) driven by the list of registered product versions (`useRegisteredProductVersions` / `registeredProductVersions` query). It's used instead of the existing `ProductVersionFilter` when the `DECOUPLING_CONNECTORS` feature flag is enabled.
- The selected version is never sent to the backend as a query filter — it's stored client-side (local storage, per service list) and only drives which connector cards get greyed out, with a tooltip explaining the incompatibility, on both the private and public paths.
- If the user's organization already has a registered platform instance for a listed version, the filter option shows an info tooltip with the matching instance name(s) (`groupInstanceNamesByVersion`).
- Backend: decoupled (V2) connectors don't carry `product_version` metadata, only `minimum_deployable_version`. The `Connector.product_version` resolver now falls back to `minimum_deployable_version` so the filter and card display keep working for those connectors too.
- `useRegisteredPlatforms` gained a `skip` option (GraphQL `@skip`) so public, unauthenticated pages don't attempt the authenticated `registeredPlatforms` query.

This PR follows up on #3205; related work in #3206 and #3207.

https://github.com/user-attachments/assets/f44b2cb9-9e31-4d64-b0f1-67c9ced90eef

Prerequisite
- Create new OpenCTI versions
- Create new connectors with different version and compatibility version with OpenCTI.
- Registered an OpenCTI versions

Once it's done:
- You can play and filter with OpenCTI version
- Greyed out with a tooltip in case is not compatible in private and public path

## Related issues

- Related to : 
- #3205 
- #3206 
- #3207

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/frontend/src/components/ui/shareable-resource/IntegrationAccordion.tsx
```diff
@@ -24,14 +24,11 @@ const IntegrationAccordion = ({
 
       defaultValue={[integrationType]}>
       <AccordionItem value={integrationType}>
-        <h2 className="m-0">
```
jpkha, line 27: remove unnecessary h2

PR #3281, opened 2026-09-03
feat(filters): Add filters facets (#3049)

hervyt:
# Context
>

This PR implements the document filters (facets) in the Hub public and private pages: a filter sidebar with per-option counts replaces the previous chips UI, backed by a new documentFacets GraphQL query.

During review, the backend layer was rewritten following a measured performance audit (see the request-changes thread): the original implementation fanned out into 8 scoped DB queries per facet request, holding 8 of the pool's 10 connections — a starvation risk on public pages. It now runs as a single SQL round-trip: one MATERIALIZED CTE per distinct stripped-filter signature, then a UNION ALL of all 8 facet aggregations. The decision between the two candidate strategies (grouped-by-signature vs single query) was settled by a 60k-doc benchmark plus a concurrent load test against a shared pool with simulated unrelated traffic: grouped was 1.4-3.1× faster in isolation, but starved unrelated queries up to 96× at 5 concurrent page loads — single query wins under any realistic concurrency. Counts are byte-identical to the previous implementation (parity-tested), including the intentionally preserved silent-operator quirk.

The facets query was also migrated from Relay to react-query/codegen (no new Relay rule); the documents query stays Relay for a follow-up. The shared facet-counts logic between the public and private lists is extracted into useDocumentFacetCounts (per review feedback).

1. Log in with an account whose organization subscribes to a private OpenCTI Integrations instance, and open its private page: the filter sidebar renders with translated titles and counts.
2. Accordions are expanded by default; each option shows its count, including draft documents for accounts that can see them.
3. Apply a filter (UI or ?licenseType=... URL param): the list and every remaining facet count refresh together.
4. Shrink the viewport until the list header wraps to two lines: the sidebar slides down and stays below it (offset tracked at runtime).
5. On a page with little content, expand all filters: the sidebar gets its own inner scroll and the last accordion remains reachable and closable.

## Related issues

- Related to #3049

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [x] Integration tests
- [ ] E2E tests
- [x] Local manual tests

# Additional information


apps/frontend/src/hooks/use-service-list-local-storage.test.tsx
```diff
@@ -79,6 +76,21 @@ describe('useServiceListLocalStorage', () => {
     }
   );
 
+  it('uses distinct storage keys for count and pageSize', () => {
```
Kroustille: What's the difference between `count` and `pageSize` ? Shouldn't it be the same values ?

apps/frontend/src/components/ui/shareable-resource/integration/IntegrationFilters.tsx
jpkha, line 1: Looks like this components is never used @Antoine-Quintin

apps/frontend/src/components/service/document/public-document.graphql.ts
```diff
@@ -271,6 +271,44 @@ export const PublicDocumentListQuery = graphql`
     ...publicDocumentList
   }
 `;
+export const FacetDocumentListQuery = graphql`
```
Kroustille: This query must be a `react-query` instead of a relay one

apps/frontend/src/components/ui/shareable-resource/PublicShareableDocumentList.tsx
```diff
@@ -101,7 +101,7 @@ export const PublicShareableDocumentList = ({
       {displayMode === ServiceListDisplayMode.Tab ? (
         <ul
           className={
-            'grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 2xl:grid-cols-4 gap-l'
+            'grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 2xl:grid-cols-3 gap-l'
```
jpkha: Make it more consistent with grid-template-columns: repeat(auto-fit, ...)
Don't forgot the private part

apps/backend/src/modules/document/facet/facet.domain.ts
```diff
@@ -0,0 +1,261 @@
...
+    const [
+      integrationType,
+      licenseType,
+      managerSupported,
+      verified,
+      productVersion,
+      solutionCategory,
+      useCase,
+      entityType,
+    ] = await Promise.all([
```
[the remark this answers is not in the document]
Kroustille: This is indeed an issue that must be addressed because it spawns 8 full table queries which uses 8 out of 10 connection in the knex pool

apps/frontend/src/components/service/components/ServiceList.tsx
```diff
@@ -23,16 +22,17 @@ import { useUserHasPortalCapability } from '@/hooks/use-portal-capability';
...
-import { Fragment, useLayoutEffect } from 'react';
+import { Fragment, useLayoutEffect, useRef } from 'react';
 
 export interface ServiceListProps {
   active: documentItem_fragment$data[];
   draft: documentItem_fragment$data[];
   search: string;
   onSearchChange: (v: string) => void;
-  additionalFilters?: ServiceListFilterMap;
+  additionalFilters: ServiceListFilterMap;
```
jpkha, line 35: does we really need additionalFilters to be mandatory ?

Antoine-Quintin: I'd keep it mandatory. ServiceList has exactly one render site ShareableResourceServiceList, which always passes filters from the shared mapping so no current caller needs it optional.

apps/frontend/src/components/service/document/PublicDocumentsList.tsx
```diff
@@ -46,6 +56,82 @@ const PublicDocumentsList = ({
...
+      input: {
+        serviceInstanceId: serviceInstance.id,
+        documentType:
+          SERVICE_SLUG_SHAREABLE_RESOURCE_MAPPING[serviceInstanceSlug],
+        searchTerm: search,
+        // useLogicalFiltersFromStorage is shared with the (still-Relay) documents
+        // query, so it's typed against Relay's LogicalFilterInput (all fields
+        // optional). The react-query codegen output for the same schema input
+        // type has `avoidOptionals` enabled, making fields required-but-nullable
+        // instead — structurally identical at runtime, just a stricter TS shape.
```
Kroustille: I think we can remove this comments

apps/frontend/src/components/service/document/PublicDocumentsList.tsx
```diff
@@ -46,6 +56,82 @@ const PublicDocumentsList = ({
...
+    },
+    { placeholderData: keepPreviousData }
+  );
+
+  // `facetData` is `undefined` until the query has resolved at least once
+  // (react-query's initial-fetch state). Returning `undefined` here — instead
+  // of computing all-zero counts — lets each filter badge stay hidden rather
+  // than briefly rendering "(0)" before the real counts arrive. Once fetched,
+  // `placeholderData: keepPreviousData` keeps `facetData` populated across
+  // refetches, so this only ever gates the very first paint.
```
Kroustille: Same here for the comments

apps/frontend/src/components/service/components/ShareableResourceServiceList.tsx
```diff
@@ -94,6 +113,68 @@ const ShareableResourceServiceList = ({
...
+      facetData === undefined
+        ? undefined
+        : toServiceListFacetCounts(facetData.documentFacets),
+    [facetData]
+  );
+
+  const { filters } = useShareableResourceMapping(
+    serviceInstanceSlug,
+    facetCounts
+  );
```
Kroustille: This part of code is duplicated with the public component. Do you think it's a good idea to mutualize it ?

apps/backend/knexfile.ts
```diff
@@ -126,6 +126,15 @@ pkg.QueryBuilder.extend('tap', function (fn) {
...
 const config: Knex.Config = {
   ...baseConfig,
+  // Scoped to this single app-wide `database` instance — NOT `baseConfig`,
+  // which is also `knex(baseConfig)`'d directly by several ES-migration
+  // scripts that never call `.destroy()`. Bumping the pool there would
+  // multiply leaked idle connections by the number of those instances.
+  // max: 20 covers the worst case of 8 concurrent facet queries (now down
+  // to a single round-trip per request) plus headroom for every other
+  // resolver sharing this pool; min: 2 avoids reconnecting on every burst
+  // after an idle period.
```
Kroustille: This comment is confusing because he talks about 8 concurrent facet queries which were never implemented (because never merged).

I think we can remove this comment entirely.

apps/backend/knexfile.ts
```diff
@@ -126,6 +126,15 @@ pkg.QueryBuilder.extend('tap', function (fn) {
...
+  pool: { min: 2, max: 20 },
```
Kroustille: I'm wondering if this change could cause issues when going to production 🤔

Antoine-Quintin: The reason for raising it to 20 was the 8-query fan-out — which never shipped. With the single query, each facet request uses 1 connection, so I'm dropping back to max: 10 (knex's default, but written explicitly so it's visible and documented). Going above 10 would need real prod numbers: instance count × pool max has to stay under Postgres's max_connections. Do we have those ?

Kroustille (approved):

PR #3339, opened 2026-09-08
fix(trials): fix redirects behavior (#3336)

theochhean:
# Context

- Back link missing `/app` prefix, `Back to XTM Platform Trial` pointed to `service/xtm-platform-trial` instead of `app/service/xtm-platform-trial`, landing on a broken route
- `Access` link resolving internally: scheme-less platform URLs (e.g. `opencti.example.com`) were treated as relative paths and redirected to `localhost:3002/app/service/...` Now normalized to external `https://` URLs
- Confusing back button from admin dashboard: opening manage-trials from the dashboard showed `Back to XTM Platform Trial`, it now shows `Back to dashboard`

## How to test

- Test that `Back to XTM Platform Trial` button redirects properly
https://github.com/user-attachments/assets/6a659428-1b84-4cb2-835e-7c712cfa827d

- Test that the `Back to ...` label adapts to where the page was reached
https://github.com/user-attachments/assets/4a8da7d2-2cf8-443e-ac8f-621ed3f860fe

## Related issues

- Related to #3336

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information

### DIFF
--- a/apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/manage-users/page.tsx
+++ b/apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/manage-users/page.tsx
@@ -7,7 +7,7 @@ import ClientSection from './client-section';
 
 export interface ServiceXtmPlatformBundleManageUsersPageProps {
   params: Promise<{ serviceInstanceId: string }>;
-  searchParams: Promise<{ from?: string }>;
+  searchParams: Promise<{ from?: string | string[] }>;
 }
 
 const Page = async ({
@@ -22,7 +22,8 @@ const Page = async ({
   }
 
   const { from } = await searchParams;
-  const fromDashboard = from === MANAGE_USERS_ORIGIN_DASHBOARD;
+  const fromValue = Array.isArray(from) ? from[0] : from;
+  const fromDashboard = fromValue === MANAGE_USERS_ORIGIN_DASHBOARD;
 
   return (
     <GuardCapacityComponent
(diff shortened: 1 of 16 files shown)


apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/manage-users/page.tsx
```diff
@@ -1,15 +1,18 @@
 import GuardCapacityComponent from '@/components/AdminGuard';
+import { MANAGE_USERS_ORIGIN_DASHBOARD } from '@/utils/path/constant';
 import { isFeatureEnabled } from '@/utils/settings.service';
 import { FeatureFlag, OrganizationCapability } from '@graphql/generated';
 import { notFound } from 'next/navigation';
 import ClientSection from './client-section';
 
 export interface ServiceXtmPlatformBundleManageUsersPageProps {
   params: Promise<{ serviceInstanceId: string }>;
+  searchParams: Promise<{ from?: string | string[] }>;
```
Comment (Julienger0) [inline apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/[serviceInstanceId]/manage-users/page.tsx]: [nit] Shouldn't we reuse the const MANAGE_USERS_ORIGIN_PARAM that you created in constant.ts to be sure that both sides stay in sync (like you did with MANAGE_USERS_ORIGIN_DASHBOARD)?

apps/frontend/src/components/xtm-platform-trial/BundleProductCard.tsx
```diff
@@ -50,7 +51,8 @@ export const BundleProductCard = ({
   const hasAccess = (registeredPlatform?.myGroups?.length ?? 0) > 0;
   const accessUrl =
     registeredPlatform?.url?.trim() || product.url?.trim() || null;
-  const canAccess = isXtmone ? !!accessUrl : hasAccess && !!accessUrl;
+  const accessHref = toExternalHref(accessUrl);
```
Julienger0, line 54: [nit] accessUrl trim() then toExternalHref trims again , maybe we can remove the first one

Julienger0 (approved):

apps/frontend/src/components/xtm-platform-trial/BundleProductCard.tsx
theochhean: we can't remove the first one as it's a fallback to determine which url to take, and because non empty whitespace strings ("     ") will result as truthy, which we want to avoid
we could remove the `toExternalHref` trim, but as it is an util and can be used elsewhere, we can't be sure that the passed arg will always be trimmed, so this acts as a guard for that case

PR #3387, opened 2026-09-11
feat(integrations): add export CSV endpoint (#3384)

jpkha:
# Context
Adds a new REST endpoint, `GET /document/csv-export/:serviceInstanceId`, that exports the Integrations Library (OpenCTI integrations) documents visible to the logged-in user as a downloadable CSV file.

- Restricted to the Integrations Library service (`ServiceDefinitionIdentifier.OpenctiIntegrations`); returns `400` for any other service instance.
- Supports an optional `columns` query param to pick which extra columns are included: `integration_type`, `use_case`, `deployment_type`, `verification_status`, `solution_category`, `license_type`, `short_description`, `long_description`, `feed_url`, `opencti_documentation`, `vendor_url`, `demo_link`; `name` is always included. Falls back to all columns when omitted/empty/unrecognized.
- Supports optional row filters (`integration_type`, `license_type`, `manager_supported`, `verified`, `solution_category`, `use_case`) narrowing which integrations are exported, reusing the existing `LogicalFilterInput`/`FilterKey` filtering used by `DocumentApp.loadDocuments`.
- `DocumentApp.loadDocuments`/`DocumentDomain.loadParentDocumentsByServiceInstance` now accept a caller-provided metadata key list (`INTEGRATION_CSV_EXPORT_METADATA_KEYS`) instead of always hydrating the full service-definition metadata set, and `first` is now optional; the CSV export omits it entirely, so it runs as a single unpaginated, uncapped query (no row limit) rather than reusing the paginated GraphQL API's page size.
- Rate-limited to 10 exports/minute/IP via `express-rate-limit`.
- Sends a telemetry export event (new `TelemetryHelper.buildExportEvent`) when telemetry is enabled for the service.
- A failed use-case/solution-category dataloader lookup for any row now fails the whole export (re-thrown), instead of silently rendering that column empty for the affected row.
- Neutralizes CSV formula injection (CWE-1236) by prefixing formula-like cells (starting with `=`, `+`, `-`, `@`, tab, or CR) with an apostrophe before RFC 4180 escaping.
- Returns a CSV file (`Content-Type: text/csv`) with a UTF-8 BOM, CRLF line endings, and a `Content-Disposition` filename of `integrations-library-export-YYYY-MM-DD.csv`.

ℹ️ **Feature environment** — A dedicated preview environment (`https://dev-pr-{number}.hub.staging.filigran.io`) is automatically deployed when CI passes. Add the `skip-feature-env` label to opt out.

## How to test
1. Start the backend (`yarn dev:api`) and log in to get a session cookie.
2. Use the `bruno/integration-csv-export/ExportIntegrationsCsv.bru` request (set `cookie` in your Bruno environment) against the Integrations Library service instance, or hit the endpoint directly:
   `GET /document/csv-export/:serviceInstanceId?columns=integration_type,use_case,deployment_type,verification_status,solution_category,license_type,short_description,long_description,feed_url,opencti_documentation,vendor_url,demo_link`
3. Verify the response is a CSV file with the requested columns and only the `name` column when `columns` is omitted.
4. Try enabling one or more of the commented-out row filters (`integration_type`, `license_type`, `manager_supported`, `verified`, `solution_category`, `use_case`) in the Bruno request and confirm the export is narrowed accordingly.
5. Hit the endpoint with a `serviceInstanceId` for a non-Integrations-Library service and confirm `400`; with an unknown one and confirm `404`; while logged out and confirm `401`.
6. Call the endpoint more than 10 times within a minute and confirm rate limiting kicks in.
7. Export an integration whose name/description starts with `=`, `+`, `-` or `@` and confirm the cell opens in Excel/Sheets as text (leading apostrophe) rather than as a formula.
8. With a large integrations catalog, confirm the export includes every row (no truncation, since there is no row cap).

## Related issues

- Related to #3384

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information
None.

### DIFF
--- a/apps/backend/src/modules/shareable-resource/opencti/integration/integration-csv-export-endpoint.ts
+++ b/apps/backend/src/modules/shareable-resource/opencti/integration/integration-csv-export-endpoint.ts
@@ -1,0 +1,183 @@
+import cors from 'cors';
+import { Application, Request, Response } from 'express';
+import rateLimit from 'express-rate-limit';
+import {
+  DocumentOrdering,
+  OrderingMode,
+  ServiceDefinitionIdentifier,
+} from '../../../../__generated__/resolvers-types';
+import { requestContext } from '../../../../context/request.context';
+import { OrganizationId } from '../../../../model/kanel/public/Organization';
+import { ServiceInstanceId } from '../../../../model/kanel/public/ServiceInstance';
+import { UserId } from '../../../../model/kanel/public/User';
+import { logApp } from '../../../../utils/app-logger.util';
+import { extractId } from '../../../../utils/utils';
+import { DocumentApp } from '../../../document/document.app';
+import { DocumentDataLoader } from '../../../document/document.dataloader';
+import { OrganizationDomain } from '../../../organization-management/organization/organization.domain';
+import { ServiceInstanceDomain } from '../../../service/instance/service-instance.domain';
+import { TelemetryApp } from '../../../telemetry/telemetry.app';
+import { TelemetryHelper } from '../../../telemetry/telemetry.helper';
+import {
+  buildIntegrationsCsv,
+  buildIntegrationsExportFilename,
+  IntegrationCsvExportRow,
+  parseRequestedColumns,
+  parseRequestedFilters,
+} from './integration-csv-export.util';
+
+// Realistic upper bound for the number of integrations in the library today;
+// keeps the export a single unpaginated query while still bounding memory.
+export const EXPORT_MAX_ROWS = 10_000;
+
+const csvExportRateLimiter = rateLimit({
+  windowMs: 60 * 1000, // 1 minute
+  max: 10, // max 10 exports per minute per IP
+  standardHeaders: true,
+  legacyHeaders: false,
+});
+
+const sendTelemetry = async (
+  serviceDefinitionIdentifier: ServiceDefinitionIdentifier,
+  userId: UserId,
+  selectedOrganizationId: OrganizationId
+): Promise<void> => {
+  try {
+    if (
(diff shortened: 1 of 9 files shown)


apps/backend/src/modules/document/document.app.ts
```diff
@@ -578,7 +578,14 @@ export const DocumentApp = {
     return documentFromDb as T;
   },
 
-  loadDocuments: async (input: QueryDocumentsArgs) => {
+  // `metadataKeysOverride` lets callers (e.g. the CSV export) hydrate a caller-chosen metadata
+  // key list instead of the service definition's default one, avoiding unused metadata. `first`
+  // is optional in the input type so exports aren't row-capped like the paginated GraphQL API.
```
Kroustille, line 583: I'm not sure this comment is needed

apps/backend/src/modules/shareable-resource/opencti/integration/integration-csv-export-endpoint.ts
jpkha, line 1: I will move the endpoint to the correct directory.

Kroustille (commented):
The PR mentions that export is capped at `EXPORT_MAX_ROWS` but I can't find it in changes 🤔

apps/backend/src/modules/document/document.app.ts
jpkha: yes sorry. I remove it. I find out it's too early to try to add a max rows. So I decide to remove it. Once we reach a certain limit we will see if we need to improve.

Kroustille (approved):

Issue #3447, opened 2026-09-16
perf(backend): sitemap document query fetches full rows and unused metadata


jpkha:
## Context

`app/sitemap.ts` calls `fetchDocumentSlugsForSitemap`, which only needs each public document's `slug`, `created_at`, and `updated_at`. That util ran the `PublicDocumentsByServiceSlugSitemapQuery` GraphQL operation, which was backed by the exact same resolver/query as the full catalog page (`publicDocumentsByServiceSlug`):

- `SELECT Document.*` — fetches all ~19 columns (including `description`/`tags`) even though only 3 are needed.
- Unconditionally runs a metadata-hydration query against `Document_Metadata`, even though the sitemap never reads hydrated metadata fields.

This runs once per public, document-bearing service instance on every sitemap generation / cache invalidation, and is pure wasted I/O.

## Proposal

Add a lightweight, dedicated GraphQL query (`publicDocumentSlugsByServiceSlug`) and backend loader (`loadSeoDocumentSlugsByServiceSlug`) that select only `slug`/`created_at`/`updated_at` and skip metadata hydration entirely, reusing the existing filtering/decoupling logic in `buildSeoDocumentsByServiceSlugQuery`. Point the sitemap's GraphQL operation and frontend fetch util at the new query.

PR #3448, opened 2026-09-16
perf(backend): add lightweight sitemap document slug query (#3447)

jpkha:
# Context
The sitemap (`app/sitemap.ts`) fetches every public document's `slug`/`created_at`/`updated_at` per service instance via `fetchDocumentSlugsForSitemap`. That util reused the exact same GraphQL resolver/query as the full catalog page (`publicDocumentsByServiceSlug`), which:
- runs `SELECT Document.*` — all ~19 columns (including `description`/`tags`), even though only 3 are needed;
- unconditionally hydrates `Document_Metadata`, even though the sitemap never reads metadata fields.

This adds a dedicated, lightweight `publicDocumentSlugsByServiceSlug` GraphQL query backed by `DocumentDomain.loadSeoDocumentSlugsByServiceSlug`. It reuses the existing filtering/decoupling logic in `buildSeoDocumentsByServiceSlugQuery` (public/active/type filters, parent-only, decoupling tag restriction) but selects only `slug`/`created_at`/`updated_at` and skips metadata hydration entirely. The sitemap's `.graphql` operation and its frontend fetch util now point at this new query. The full catalog page's query/behavior is unchanged.

ℹ️ **Feature environment** — A dedicated preview environment (`https://dev-pr-{number}.hub.staging.filigran.io`) is automatically deployed when CI passes. Add the `skip-feature-env` label to opt out.

## How to test
1. Start the API and frontend (`yarn dev:api`, `yarn dev:front`).
2. Visit `/en/cybersecurity-solutions/<service-slug>` and confirm the catalog page still lists all public documents unchanged.
3. Visit `/sitemap.xml` and confirm document URLs for each service instance are still listed with correct `lastModified` dates.
4. Optionally inspect backend logs/SQL to confirm the sitemap path no longer joins `Document_Metadata` or selects unused columns.

## Related issues

- Related to #3447

# Checklist

## Quality

## Testing

### What tests have been made

- [x] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information
Backend `check-ts`/lint and frontend `check-ts`/lint/unit tests all pass. New backend domain/app/resolver tests mirror the existing `loadSeoDocumentsByServiceSlug`/`loadPublicDocumentsByServiceSlug` test suites but could not be executed here — this sandbox has no local Postgres/docker to run the backend's DB-backed integration tests; please confirm they pass in CI.


apps/backend/src/modules/document/domain/document.domain.test.ts
```diff
@@ -1015,6 +1015,106 @@ describe('document domain', () => {
...
 
+  describe('loadSeoDocumentSlugsByServiceSlug', () => {
+    const TEST_SERVICE_SLUG = 'opencti-integrations';
+
+    let parentDoc: Document;
+    let childDoc: Document;
+    let inactiveDoc: Document;
+    let otherServiceDoc: Document;
+
+    beforeEach(async () => {
```
carinelebas, line 1026: Since we are just testing a 'read' function, maybe we could use a beforeAll here?

carinelebas (approved):

apps/backend/src/modules/document/domain/document.domain.test.ts
jpkha: there is other test inside this document with creating Document. If we use beforeAll it may create some conflict with other

Issue #3414, opened 2026-09-14
feat(xtm trial): Ability to send and email when adding a user


elisavaldegas:
Send an email to the added user when an Admin adds a user to a trial

PR #3451, opened 2026-09-16
feat(trials): send an XTM Platform invitation mail when adding a user to a bundle trial (#3414)

Julienger0:
# Context
>

## Related issues

- Related to #3414

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/server/mail-service.helper.test.ts
```diff
@@ -697,6 +697,50 @@ describe('free_trial_bundle_active bullets', () => {
   });
 });
 
+describe('free_trial_bundle_user_added', () => {
+  afterEach(() => {
+    clearTemplateCache();
+  });
+
+  it('should render the invitation with the admin, the products, the days left and the links', async () => {
+    const platformUrl = 'https://hub.filigran.io/app/xtm-platform-trial';
```
carinelebas: is not the url `https://hub.filigran.io/app/service/xtm-platform-trial` ?

apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -121,22 +126,37 @@ export const ServiceGroupApp = {
...
 
-    await withTransaction(async () => {
-      for (const { child, role } of platformRoleAssignments) {
-        const groups = await ServiceGroupDomain.loadServiceGroups({
-          service_instance_id: child.service_instance_id,
-        });
-        const targetGroup = groups.find((group) => group.name === role);
-        if (!targetGroup) {
-          throw new Error(ErrorCode.ServiceGroupNotFound);
+    const existingBundleUserIds = await withAdvisoryLock(
```
carinelebas: Having the same user added twice at the same time is really an edge case. I'm not sure it's worth adding the complexity. At worse, we'll send 2 mails, this is not a critical issue, and it most likely will never happen.

apps/backend/src/modules/deployment/group/service-group.app.test.ts
```diff
@@ -641,6 +642,50 @@ describe('serviceGroupApp', () => {
       await expect(call).rejects.toThrow(ErrorCode.XtmOneRoleRequired);
     });
 
+    it('should keep only the first role when the same product is submitted twice', async () => {
```
carinelebas, line 645: when can this case happen?

apps/backend/src/server/mail-service.helper.test.ts
Julienger0: True, It was also wrong in one of my former changes in another PR, I fixed it as well

apps/backend/src/modules/deployment/group/service-group.app.test.ts
Julienger0: I deleted it, you're right it's notpossible from the UI

apps/backend/src/modules/deployment/group/service-group.app.ts
```diff
@@ -121,10 +121,16 @@ export const ServiceGroupApp = {
 
     const platformRoleAssignments = ServiceGroupHelper.matchRolesToChildren(
       children,
-      input.roles
+      ServiceGroupHelper.uniqueRolesByProduct(input.roles)
     );
 
-    await withTransaction(async () => {
+    const existingBundleUserIds = await withTransaction(async () => {
+      const existingUserIds =
```
carinelebas: I don't think this is needed: this function is normally called only for users that don't already exist in the table. 
If we really want to protect anyway, this is to update ServiceGroupDomain.addUsersToGroup so it returns the data inserted (the .onConflict.ignore will return only the data really inserted, not the one in conflict), and send the mail only for these users. It would actually also cover the edge case where the users are concurrently inserted.

PR #3466, opened 2026-09-17
feat(xtmplatformtrial-14): cleanup & FF removal (#3313)

Julienger0:
# Context
>
What was removed:

Feature flag
- `XTM_PLATFORM_TRIAL` — the XTM Platform trial pages, banners and menu links are now always on.

Standalone trial creation
- The "Start free trial" pages (`/app/service/opencti-free-trial`, `/app/service/openaev-free-trial`) and their public counterparts, with their request form.
- The "Try Filigran products" banners (public and private) and the "Try OpenCTI / OpenAEV" block on the homepage.
- The per-product "Start free trial" links in the private and public menus.
- The OpenCTI / OpenAEV deep links `create-free-trial` and `create-openaev-free-trial` now land on the XTM Platform trial page. Old free-trial URLs redirect there too.
- Backend: the create mutation only accepts bundles (the `type` field is gone from the input), and the `trialDeployments` query.
- Emails sent when a standalone trial was requested or queued, and the matching admin notification.

Product quotas
- Quotas are now managed per region for the bundle only; the per-product (OpenCTI / OpenAEV) quotas are gone from the code and the database (migration included).
- The "Quotas" tab is only available on *Manage trials*; the OpenCTI / OpenAEV trial lists in the admin are kept so remaining standalone trials can still be followed and cancelled.

## Related issues

- Related to #3313

# Checklist

## Quality

## Testing

### What tests have been made

- [ ] Unit tests
- [ ] Integration tests
- [ ] E2E tests
- [ ] Local manual tests

# Additional information


apps/backend/src/migrations/20260916142052_remove_product_deployment_quota.js
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  const heldPlacesByRegion = await countRootRequestsByRegion(knex);
+  const quotas = await knex('DeploymentRequestQuota').select(
+    'id',
+    'region',
+    'capacity'
+  );
+
+  for (const quota of quotas) {
```
carinelebas: question: why is it needed? is the bundle quota availability not good?

apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/page.tsx
```diff
@@ -1,17 +1,5 @@
...
-  const xtmPlatformTrialEnabled = await isFeatureEnabled(
-    FeatureFlag.XtmPlatformTrial
-  );
-  if (!xtmPlatformTrialEnabled) {
-    notFound();
-  }
-
-  return <PageLoader />;
-};
+const Page = () => <PageLoader />;
```
carinelebas, line 3: just wondering: do we really need to keep this file then?

apps/backend/src/modules/deployment/deployment.domain.test.ts
```diff
@@ -1282,6 +1282,11 @@ describe('deploymentRequestDomain', () => {
     });
   });
   describe('loadLastPendingRequest and setRequestAsQueued', () => {
+    const BUNDLE_REQUEST = {
+      type: DeploymentRequestDeploymentType.Bundle,
+      platform_identifier: null,
```
carinelebas: question : is platform_identifier needed here? we'll always have bundles pending or set as queued, so platform_idenfier will always be null ?

carinelebas (commented):
great clean up! 🧹 Thanks for the commit split!

apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/page.tsx
Julienger0: we use it here https://hub.filigran.io/app/service/xtm-platform-trial

apps/backend/src/migrations/20260916142052_remove_product_deployment_quota.js
Julienger0: I added it because of copilot review : "This rollback reconstructs both product quota rows from the single bundle quota capacity, but up has already deleted the original per-product capacities. If OpenCTI and OpenAEV capacities differed (for example after an admin edit), down silently restores incorrect quotas and cannot return the database to its pre-migration state. Preserve the original product values or make the migration explicitly irreversible rather than fabricating them." but it's true that since the former migration was live before we activated the quota we don't need it, I removed it

apps/backend/src/modules/deployment/deployment.domain.test.ts
Julienger0: No you're right, I'll take it off

apps/frontend/app/(application)/app/(user)/service/xtm-platform-trial/page.tsx
carinelebas: I was mainly thinking that now it's just a passthrough to pageLoader, so maybe we could merge both. But since PageLoader is a client component, maybe it's better to keep both.. 🤔

carinelebas (approved):

apps/backend/src/migrations/20260916142052_remove_product_deployment_quota.js
carinelebas: I think it's safer to avoid updating something that works 😄

